PR #19, opened 2015-01-09
DocForm remove obj

msladek:
- added functionality to remove obj via DocFormCommand (with preceding "^" to objNb)
- refactored request key parsing for DocFormCommand to own parser with return class

### DIFF
--- a/src/main/java/com/celements/docform/DocFormRequestKey.java
+++ b/src/main/java/com/celements/docform/DocFormRequestKey.java
@@ -47,8 +47,6 @@ public class DocFormRequestKey {
   }
 
   public boolean sameObject(DocFormRequestKey other) {
-    System.out.println("compare " + this);
-    System.out.println("with " + other);
     return (this.classRef != null) && (this.objNb != null) && (this.objNb >= 0)
         && ObjectUtils.equals(this.classRef, other.classRef) 
         && ObjectUtils.equals(this.objNb, other.objNb);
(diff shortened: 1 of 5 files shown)


src/main/java/com/celements/docform/DocFormRequestKey.java
```diff
@@ -0,0 +1,64 @@
...
+    return objNb;
+  }
+
+  public String getFieldName() {
+    return fieldName;
+  }
+
+  public boolean sameObject(DocFormRequestKey other) {
+    System.out.println("compare " + this);
+    System.out.println("with " + other);
```
ebeutler: Please change to log or remove.

src/test/java/com/celements/web/plugin/cmd/DocFormCommandTest.java
```diff
@@ -122,67 +87,123 @@ public void testGetUpdateDoc() throws XWikiException {
...
-    verifyAll(obj);
+    
+    replayDefault();
+    docFormCmd.setOrRemoveObj(docMock, key, value, context);
+    verifyDefault();
+    
+    assertEquals(objMock, docFormCmd.getChangedObjects().get("W:Sp.Doc_W:A.B_0"));
   }
   
+  // TODO do nothing in this case?
```
ebeutler: Still needed?

ebeutler:
Two small things, rest looks fine.

fpichler:
Please do NOT remove any public methods from ScriptService classes (e.g. CelementsWebScriptService) instead mark them deprecated and add a delegation to the new ScriptService method. Otherwise all velocity scripts using this public methods will brake!

fpichler:
Please do NOT use commons-logging any more. Thus replace all commons-LOGGER in new classes with slf4j logger.

fpichler:
sorry, wrong pull request for the last to messages. :-(

PR #23, opened 2015-05-20
Feature paging nav

ebeutler:


fpichler:
add missing unit tests and fixed implementation. Can you please review.

msladek:
Two questions to consider from my part:
1. offset isn't considered at all if nrOfItemsPerPage isn't set. Is this intentional? Can't/shouldn't the client be able to use offset on it's own?
2. Wouldn't it be less error-prone to treat nrOfItemsPerPage =  0 like -1? I can't imagine a sensible use case where one would want to return 0 items.

fpichler:
- added support for offset without paging
- handling of "nrOfItemsPerPage = 0 like -1" has been there already
Please code review and merge if possible?

msladek:
yes you are right, its already in there. Looks good now to me.

PR #25, opened 2015-06-15
Classes isActivated & Scheduler class collection

msladek:


src/main/java/com/celements/scheduler/classes/SchedulerClasses.java
```diff
@@ -0,0 +1,84 @@
...
+  public String getConfigName() {
+    return "scheduler";
+  }
+
+  @Override
+  protected void initClasses() throws XWikiException {
+    getSchedulerJobClass();
+  }
+
+  protected BaseClass getSchedulerJobClass() throws XWikiException {
```
fpichler: please change to private (or package private)

PR #26, opened 2015-06-23
NavigationParents

msladek:
ticket https://synjira.atlassian.net/browse/CELDEV-42

### DIFF
--- a/src/main/java/com/celements/parents/DocumentParentsLister.java
+++ b/src/main/java/com/celements/parents/DocumentParentsLister.java
@@ -2,8 +2,11 @@ package com.celements.parents;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -11,7 +14,12 @@ import org.xwiki.component.annotation.Component;
 import org.xwiki.component.annotation.Requirement;
 import org.xwiki.model.reference.DocumentReference;
 
+import com.celements.pagetype.IPageTypeConfig;
+import com.celements.pagetype.IPageTypeProviderRole;
+import com.celements.pagetype.service.IPageTypeResolverRole;
+import com.celements.pagetype.xobject.XObjectPageTypeProvider;
 import com.celements.web.service.XDocRecursionException;
+import com.google.common.collect.Iterables;
 
 @Component
 public class DocumentParentsLister implements IDocumentParentsListerRole {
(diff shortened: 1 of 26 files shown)


src/main/java/com/celements/parents/DocumentParentsLister.java
```diff
@@ -21,61 +29,94 @@
...
 
-  private void joinParentLists(ArrayList<DocumentReference> docParents,
-      List<DocumentReference> parentList) throws XDocRecursionException {
-    for (DocumentReference parentRef : parentList) {
-      if (!docParents.contains(parentRef)) {
-        docParents.add(parentRef);
-      } else {
-        throw new XDocRecursionException(parentRef);
+  private Set<String> getProvidersWithout(String provider) {
+    Set<String> ret = new HashSet<>(docParentProviderMap.keySet());
```
ebeutler: I usually repeat the <String> even if not strictly necessary. Do we have a code style / best practice for that? If not maybe we should discus how to handle these cases.

src/main/java/com/celements/parents/DocumentParentsLister.java
```diff
@@ -21,61 +29,94 @@
...
+    ret.remove(XDocParents.DOC_PROVIDER_NAME);
+    ret.remove(provider);
+    return ret;
+  }
+
+  private List<DocumentReference> checkPageTypes(List<DocumentReference> parents) {
+    List<DocumentReference> ret = new ArrayList<>();
```
ebeutler: see comment above.

src/main/java/com/celements/web/service/WebUtilsService.java
```diff
@@ -1459,4 +1461,11 @@ public EntityType resolveEntityTypeForFullName(String fullName,
     return componentManager.lookupMap(role);
   }
 
+  private IDocumentParentsListerRole getDocumentParentsLister() {
+    if (docParentsLister == null) {
+      docParentsLister = Utils.getComponent(IDocumentParentsListerRole.class);
```
ebeutler: Is the problem with the cyclic dependency resolved? Otherwise please add the comment again.

ebeutler:
A few small comments, looks ok apart from that.

src/main/java/com/celements/parents/DocumentParentsLister.java
msladek: I don't think we have ever discussed a code style regarding automatic type inference since it is a Java 7 feature and we only recently switched to Java 7.

src/main/java/com/celements/web/service/WebUtilsService.java
msladek: I refactored the component to a class field and therefore placed the comment there:

/*
 * not loaded as requirement due to cyclic dependency
 */
IDocumentParentsListerRole docParentsLister;

PR #31, opened 2016-01-07
fix bug where an upload of a file where the name is the prefix of an

ebeutler:
existing file leads to an overwrite of said file.


fpichler:
If doc.getAttachment does not correctly check for the filename we have likely a second bug on line 426 in deleteAttachmentMap method. Can you please verify this?

Furthermore I suggest to add a getAttachment(XWikiDocument doc, String filename) api method to IModelAccessFacade which wrappes and fixes the doc.getAttachment method. After that please use the IModelAccessFacade in AttachmentService to access the Attachments.

PR #33, opened 2016-03-29
add method getAttachmentFirstNameMatch to AttachmentService

ebeutler:


msladek:
looks good

src/main/java/com/celements/filebase/AttachmentService.java
```diff
@@ -500,6 +508,16 @@ public XWikiAttachment getAttachmentNameEqual(AttachmentReference attachmentRef
     XWikiDocument doc = modelAccess.getDocument(docRef);
     return getAttachmentNameEqual(doc, attachmentRef.getName());
   }
+  
+  @Override
+  public XWikiAttachment getAttachmentFirstNameMatch(XWikiDocument document,
+      IAttachmentMatcher attMatcher) throws AttachmentNotExistsException {
+    List<XWikiAttachment> attList =  getAttachmentsNameMatch(document, attMatcher);
+    if(!attList.isEmpty()) {
+      return attList.get(0);
```
fpichler: why would you want to get randomly the first element?

src/main/java/com/celements/web/FileAction.java
```diff
@@ -312,8 +317,11 @@ public boolean renderFileFromObjectField(String filename, XWikiDocument doc,
   public boolean renderFileFromAttachment(String filename, XWikiDocument doc,
       XWikiContext context) throws IOException, XWikiException {
     _Logger.debug("... as attachment");
-    XWikiAttachment attachment = doc.getAttachment(filename);
-    if (attachment != null) {
+    try {
+      IAttFileNameMatcherRole matcher = Utils.getComponent(IAttFileNameMatcherRole.class);
```
fpichler: this will result in an ComponentLookup Exception, because there is no "default" component.

src/main/java/com/celements/web/FileAction.java
```diff
@@ -312,8 +317,11 @@ public boolean renderFileFromObjectField(String filename, XWikiDocument doc,
...
+      matcher.setFileNamePattern(filename);
+      XWikiAttachment attachment = getAttachmentService().getAttachmentFirstNameMatch(doc,
```
fpichler: this replacement looks wrong to me, because the doc.getAttachment FIRST checks for name equals and then only returns similar results. Yet, getAttachmentFirstNameMatch may return a near match eventhough there would be an exact filename match attachment.

src/test/java/com/celements/filebase/AttachmentServiceTest.java
```diff
@@ -107,35 +108,39 @@ public void testDeleteAttachmentList_list() throws XWikiException {
     attList.add(new AttachmentReference(name1, docRef));
     attList.add(new AttachmentReference(name2, docRef));
     attList.add(new AttachmentReference(name3, docRef));
-    XWikiDocument doc = createMockAndAddToDefault(XWikiDocument.class);
+    XWikiDocument doc = new XWikiDocument(docRef);//createMockAndAddToDefault(XWikiDocument.class);
```
fpichler: please remove old commented code or add a notice why it is still in the code.

fpichler:
not good for me. :-/ see comments.

ebeutler:
changed getAttachmentFirstNameMatch to prioritize exact match

fpichler:
now, I think it's ok. I merge the branch...

PR #40, opened 2016-05-16
add concurrent and immutable support to

fpichler:
https://synjira.atlassian.net/browse/CELDEV-237

### DIFF
--- a/src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
+++ b/src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
@@ -26,8 +26,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.xwiki.context.Execution;
 import org.xwiki.model.reference.DocumentReference;
 
@@ -43,7 +43,7 @@ import com.xpn.xwiki.web.Utils;
 
 public class GetNotMappedMenuItemsForParentCommand {
 
-  private static Log LOGGER = LogFactory.getFactory().getInstance(
+  private static Logger LOGGER = LoggerFactory.getLogger(
       GetNotMappedMenuItemsForParentCommand.class);
 
   private Map<String, Map<String, List<TreeNode>>> menuItems;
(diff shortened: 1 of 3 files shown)


src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
```diff
@@ -24,42 +24,60 @@
...
   private int queryCount = 0;
 
+  private XWikiContext getContext() {
+    return (XWikiContext) Utils.getComponent(Execution.class).getContext(
+        ).getProperty("xwikicontext");
+  }
+
   public GetNotMappedMenuItemsForParentCommand() {
-    menuItems = new HashMap<String, Map<String, List<TreeNode>>>();
+    menuItems = new ConcurrentHashMap<String, Map<String, List<TreeNode>>>();
```
msladek: why are you using a ConcurrentHashMap if it is only accessed in synchronized blocks? You may as well be using a HashMap or if possible refactor the code to make use of _putIfAbsent_.

src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
```diff
@@ -79,8 +97,7 @@ public GetNotMappedMenuItemsForParentCommand() {
           docCount++;
           LOGGER.debug("got item from db: " + docData[0].toString());
           oldParentKey = parentKey;
-          parentKey = getParentKey(wikiName, docData[2].toString(), docData[1].toString(),
-              context);
+          parentKey = getParentKey(wikiName, docData[2].toString(), docData[1].toString());
           if(!oldParentKey.equals(parentKey) || (menu == null)) {
```
msladek: space missing after _if_, consider using save actions ;)

src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
```diff
@@ -24,42 +24,60 @@
...
 import com.xpn.xwiki.XWikiContext;
 import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.doc.XWikiDocument;
 import com.xpn.xwiki.objects.BaseObject;
+import com.xpn.xwiki.web.Utils;
 
 public class GetNotMappedMenuItemsForParentCommand {
 
   private static Log LOGGER = LogFactory.getFactory().getInstance(
       GetNotMappedMenuItemsForParentCommand.class);
```
msladek: Please change to slf4j when already touching class

fpichler:
Thanks for your valuable hints. I reworked the concurrency to allow unsynchronised reads. Please review again.

src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
```diff
@@ -35,31 +37,64 @@
...
+  public List<TreeNode> getTreeNodesForParentKey(String searchParentKey) {
+    String wikiName = getWikiName(searchParentKey);
+    LOGGER.trace("getNotMappedMenuItemsFromDatabase: for cacheKey [{}].", wikiName);
+    if (!menuItems.containsKey(wikiName)) {
+      loadMenuForWiki(wikiName);
+    }
+    if (menuItems.containsKey(wikiName) && (menuItems.get(wikiName) != null)
+        && (menuItems.get(wikiName).get(searchParentKey) != null)){
+      return menuItems.get(wikiName).get(searchParentKey);
+    }
```
msladek: This if block is not thread safe. If `loadMenuForWiki` or `flushMenuItemCache` are called by other threads in between these statements, data inconsistency or even NPEs could occur.

I would suggest the following changes:
```
    Map<String, List<TreeNode>> wikiMenu = menuItems.get(wikiName);
    if (wikiMenu == null) {
      wikiMenu = loadMenuForWiki(wikiName);
    }
    if (wikiMenu.containsKey(searchParentKey)) {
      return wikiMenu.get(searchParentKey);
    }
```

src/main/java/com/celements/navigation/cmd/GetNotMappedMenuItemsForParentCommand.java
```diff
@@ -35,31 +37,64 @@
...
+
   public GetNotMappedMenuItemsForParentCommand() {
-    menuItems = new HashMap<String, Map<String, List<TreeNode>>>();
+    /* ConcurrentHashMap is needed to prevent:
+     * 1) ConcurrentModificationException if Thread A iterates over menuItems and
+     *    Thread B is changing it.
+     * 2) Thread B changes the HashMap and Thread B does not synchronize on read
+     *    which may lead that it reads an old version from its memory cache.
+     *    Details see JSR133
+     */
```
msladek: 1) menuItems is never being iterated, however accessing menuItems is sufficient as a requirement for using ConcurrentHashMap
2) The second thread should be called Thread A (2x B in text)

fpichler:
fixed mentioned issues. Please review again and merge if ok.

msladek:
looks good now, thanks for the changes!

PR #41, opened 2016-05-20
Feature XObjectField

msladek:
please give thoughts on general idea and implementation.

implementation of class XObjectField which encapsulates classRef + fieldName and XObjectFieldValue with additional value for more a practical representation. first usage in ModelAccessFacade.get/setProperty and XObjectUpdater

in the future, we can extend ClassConfigs with XFields definitions to access fields easily and type safe via model, e.g. `Date myDate = model.getProperty(doc, classConf.getXFieldMyDate())`


msladek:
moved to [Pull Request 51](https://github.com/celements/celements-core/pull/51)

PR #43, opened 2016-05-27
Refactoring create new user name

ebeutler:
merging #42 first will remove part of the changes here


fpichler:
For what do you need getNewRandomXWikiUserName? Can you please explain the motivation in a CELDEV issue?

PR #45, opened 2016-06-02
add created user name to parameter map

ebeutler:
Easiest solution to return the newly created user name. Works without the need to change any of the APIs

### DIFF
--- a/src/main/java/com/celements/web/service/CelementsWebService.java
+++ b/src/main/java/com/celements/web/service/CelementsWebService.java
@@ -7,6 +7,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import javax.validation.constraints.NotNull;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xwiki.component.annotation.Component;
@@ -99,8 +101,8 @@ public class CelementsWebService implements ICelementsWebServiceRole {
   }
 
   @Override
-  public synchronized XWikiUser createNewUser(Map<String, String> userData, String possibleLogins,
-      boolean validate) throws UserCreateException {
+  public synchronized @NotNull XWikiUser createNewUser(@NotNull Map<String, String> userData,
+      @NotNull String possibleLogins, @NotNull boolean validate) throws UserCreateException {
     String accountName = "";
     String accountFullName = null;
     if (userData.containsKey("xwikiname")) {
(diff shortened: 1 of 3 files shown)


fpichler:
why not adding a new API and deprecating the old one. I do not very much fancy the idea of modifying a parameter map. At least please mention explicitly in Java Doc what can be added and removed from the map AND that it MUST be mutable.

ebeutler:
Changed implementation to add new method. Also changed some wiki calls and added not nullable annotations.

src/main/java/com/celements/web/service/CelementsWebService.java
```diff
@@ -74,60 +85,92 @@ public int createUser(boolean validate) throws XWikiException {
...
+        throw new XWikiException(XWikiException.MODULE_XWIKI_STORE,
+            XWikiException.ERROR_XWIKI_UNKNOWN, "failed to create user", uce);
+      }
+    }
+    return -1;
+  }
+
+  @Override
+  public synchronized @NotNull XWikiUser createNewUser(@NotNull Map<String, String> userData,
+      @NotNull String possibleLogins, @NotNull boolean validate) throws UserCreateException {
```
fpichler: parameter validate cannot be null because it is a basic type. Thus annotation NotNull is not needed

src/main/java/com/celements/web/service/CelementsWebService.java
```diff
@@ -74,60 +85,92 @@ public int createUser(boolean validate) throws XWikiException {
...
-              doc.getDocumentReference()));
-          rightObj.set("allow", "1", getContext());
-          rightObj.set("levels", "view,edit,delete", getContext());
-          rightObj.set("groups", "", getContext());
-        } else {
-          rightObj.set("users", "", getContext());
-          rightObj.set("allow", "1", getContext());
-          rightObj.set("levels", "view,edit,delete", getContext());
-          rightObj.set("groups", "XWiki.XWikiAdminGroup", getContext());
+      try {
```
fpichler: please extract to method "setRightsOnUserDoc"

fpichler:
oh, a lot better :+1: 
Some smaller issues, after that it is ok for me to merge.

src/main/java/com/celements/web/service/CelementsWebService.java
ebeutler: I just added @NotNull to validate to be consistent, but I think it makes sense to add those annotations for nullable types only.

PR #52, opened 2016-06-10
Fix tree node equals

fpichler:
CELDEV-257

### DIFF
--- a/src/main/java/com/celements/navigation/TreeNode.java
+++ b/src/main/java/com/celements/navigation/TreeNode.java
@@ -19,6 +19,8 @@
  */
 package com.celements.navigation;
 
+import javax.validation.constraints.NotNull;
+
 import org.apache.commons.lang.StringUtils;
 import org.xwiki.context.Execution;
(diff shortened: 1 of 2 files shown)


src/main/java/com/celements/navigation/TreeNode.java
```diff
@@ -156,19 +156,23 @@ public boolean equals(Object obj) {
...
-    if ((obj == null) || (obj.getClass() != this.getClass())) {
+    if (!(obj instanceof TreeNode)) { // null check included
       return false;
     }
     // object must be Test at this point
     TreeNode node = (TreeNode) obj;
-    return docRef.equals(node.docRef) && (position == node.position);
+    if (docRef.equals(node.docRef)) {
+      return (((position == null) && (node.position == null)) || ((position != null)
+          && position.equals(node.position)));
```
msladek: There is a neat method `java.util.Objects.equals(a, b)` (since Java 7) which handles the null checks as well.

PR #56, opened 2016-06-18
CELDEV-267: replace 'putIfAbsent' with 'put' because it is in a synchronized blog

fpichler:
https://synjira.atlassian.net/browse/CELDEV-267


msladek:
There is a [not contains check](https://github.com/celements/celements-core/blob/d4bfefc13975bb65f6818f377179fa2913f1c948/src/main/java/com/celements/pagetype/xobject/XObjectPageTypeCache.java#L99) before the _putIfAbsent_ call, therefore the key will always be absent at this point. Changing it to _put_ certainly isn't wrong, but there is no bug since the behaviour of _putPageTypeRefsForWikiToCache_ was correct before.

fpichler:
can be merged? right?

PR #57, opened 2016-06-20
ModelUtils

msladek:
https://synjira.atlassian.net/browse/CELDEV-268

### DIFF
--- a/src/main/java/com/celements/model/util/ModelUtils.java
+++ b/src/main/java/com/celements/model/util/ModelUtils.java
@@ -59,12 +59,9 @@ public final class ModelUtils {
       reference = reference.clone();
       T ret;
       if (token != EntityReference.class) {
-        ret = token.getConstructor(EntityReference.class).newInstance(reference.clone());
+        ret = token.getConstructor(EntityReference.class).newInstance(reference);
       } else {
-        // T == EntityReference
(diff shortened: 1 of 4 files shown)


src/main/java/com/celements/model/util/ModelUtils.java
```diff
@@ -0,0 +1,75 @@
...
+   * @return a cloned instance of the reference
+   */
+  @NotNull
+  public static <T extends EntityReference> T cloneReference(@NotNull EntityReference reference,
+      @NotNull Class<T> token) {
+    try {
+      reference = reference.clone();
+      T ret;
+      if (token != EntityReference.class) {
+        ret = token.getConstructor(EntityReference.class).newInstance(reference.clone());
```
fpichler: why cloning the reference twice (line 59 and line 62)?

src/main/java/com/celements/model/util/ModelUtils.java
```diff
@@ -0,0 +1,75 @@
...
+      } else {
+        // T == EntityReference
+        @SuppressWarnings("unchecked")
+        T castedReference = (T) reference;
```
fpichler: use token.cast(reference) instead '"unchecked" (T)reference'

msladek:
valid points thanks, I fixed those

PR #58, opened 2016-06-20
DocumentAccessRuntimeException

msladek:
https://synjira.atlassian.net/browse/CELDEV-269


fpichler:
On deleting a document the client must expect and cope errors, the same is true for saving a document. Moreover the client wants to know if the action was successful. Thus please do not change the DocumentDeleteException and the DocumentSaveException to runtime exceptions.

PR #63, opened 2016-07-13
Model getTranslation

msladek:
https://synjira.atlassian.net/browse/CELDEV-294


src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -66,43 +65,17 @@ private XWikiContext getContext() {
 
   @Override
   public XWikiDocument getDocument(DocumentReference docRef) throws DocumentNotExistsException {
-    checkNotNull(docRef);
-    if (exists(docRef)) {
-      return getDocumentInternal(docRef);
-    } else {
-      throw new DocumentNotExistsException(docRef);
-    }
+    return cloneDoc(getDocumentReadOnly(docRef));
```
fpichler: Please add TODO that this method should delegate to getDocument(docRef, lang) as soon as a prober ModelAccessStub is in place for tests.

PR #64, opened 2016-07-28
Concurrent cache issue

fpichler:
https://synjira.atlassian.net/browse/CELDEV-291


src/main/java/com/celements/store/DocumentCacheStore.java
```diff
@@ -0,0 +1,977 @@
...
+      if (!this.key.equals(key)) {
+        throw new RuntimeException(
+            "DocumentLoader illegally used with a different key (registered key:" + this.key
+                + ", loading doc key: " + key + ").");
+      }
+    }
+
+    private class DocumentBuilder {
+
+      private synchronized XWikiDocument buildDocument(String key, XWikiDocument doc,
```
msladek: this method needn't be synchronized, it is only used in _DocumentLoader.loadDocument_ where a new instance is created for every method call anyway

PR #70, opened 2016-08-24
DocumentMetaData

msladek:
https://synjira.atlassian.net/browse/CELDEV-315


fpichler:
thanks for the changes looks good to me.

PR #72, opened 2016-09-01
Relative ClassRef

msladek:
https://synjira.atlassian.net/browse/CELDEV-325

NOTE: These changes break the already released ClassDefinition/ClassField-API. It seems justified to me since this is a new API and these changes came up with it's first real world usage. We want to get it right before using it.

### DIFF
--- a/src/main/java/com/celements/model/classes/fields/AbstractClassField.java
+++ b/src/main/java/com/celements/model/classes/fields/AbstractClassField.java
@@ -5,10 +5,8 @@ import java.util.Objects;
 import javax.annotation.Nullable;
 import javax.validation.constraints.NotNull;
 
-import org.xwiki.model.reference.DocumentReference;
-
+import com.celements.model.classes.ClassDefinition;
 import com.celements.model.util.ModelUtils;
-import com.celements.model.util.References;
 import com.google.common.base.Strings;
 import com.xpn.xwiki.objects.PropertyInterface;
 import com.xpn.xwiki.objects.classes.PropertyClass;
@@ -19,7 +17,7 @@ import com.xpn.xwiki.web.Utils;
  */
 public abstract class AbstractClassField<T> implements ClassField<T> {
 
-  private final DocumentReference classRef;
+  private final String className;
   private final String name;
   private final String prettyName;
   private final String validationRegExp;
@@ -27,15 +25,14 @@ public abstract class AbstractClassField<T> implements ClassField<T> {
 
   public abstract static class Builder<B extends Builder<B, T>, T> {
 
-    private final DocumentReference classRef;
+    private final String className;
     private final String name;
     private String prettyName;
     private String validationRegExp;
     private String validationMessage;
 
-    public Builder(@NotNull DocumentReference classRef, @NotNull String name) {
-      Objects.requireNonNull(classRef);
-      this.classRef = References.cloneRef(classRef, DocumentReference.class);
+    public Builder(@NotNull String className, @NotNull String name) {
+      this.className = Objects.requireNonNull(className);
       this.name = Objects.requireNonNull(Strings.emptyToNull(name));
     }
 
@@ -61,7 +58,7 @@ public abstract class AbstractClassField<T> implements ClassField<T> {
   }
 
   protected AbstractClassField(@NotNull Builder<?, T> builder) {
-    this.classRef = builder.classRef;
+    this.className = builder.className;
     this.name = builder.name;
     this.prettyName = builder.prettyName;
     this.validationRegExp = builder.validationRegExp;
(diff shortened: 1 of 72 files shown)


src/test/java/com/celements/model/classes/fields/DateFieldTest.java
```diff
@@ -22,8 +22,8 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/EntityReferenceFieldTest.java
```diff
@@ -27,8 +28,7 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/EnumListFieldTest.java
```diff
@@ -33,8 +34,7 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/ListFieldTest.java
```diff
@@ -40,9 +41,8 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/NumberFieldTest.java
```diff
@@ -29,8 +29,7 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/StringFieldTest.java
```diff
@@ -20,8 +20,7 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/util/ClassFieldValueTest.java
```diff
@@ -18,14 +18,14 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/BooleanFieldTest.java
```diff
@@ -21,8 +21,8 @@
   @Before
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/test/java/com/celements/model/classes/fields/ClassFieldTest.java
```diff
@@ -28,7 +30,7 @@
   public void setUp() throws Exception {
     super.setUp();
```
fpichler: please do not override the default setUp method. Instead add a different named method with Before annotation.

src/main/java/com/celements/model/classes/fields/AbstractClassField.java
```diff
@@ -19,23 +17,22 @@
...
+    private final String className;
     private final String name;
     private String prettyName;
     private String validationRegExp;
     private String validationMessage;
 
-    public Builder(@NotNull DocumentReference classRef, @NotNull String name) {
-      Objects.requireNonNull(classRef);
-      this.classRef = References.cloneRef(classRef, DocumentReference.class);
+    public Builder(@NotNull String className, @NotNull String name) {
```
fpichler: please do not use String representations for document full names. We are working about 10 years to get rid of all String typed full names. Use DocumentReferences or EntityReferences instead.

src/main/java/com/celements/model/classes/fields/BooleanField.java
```diff
@@ -20,8 +18,8 @@
     private String displayType;
     private Integer defaultValue;
 
-    public Builder(@NotNull DocumentReference classRef, @NotNull String name) {
-      super(classRef, name);
+    public Builder(@NotNull String className, @NotNull String name) {
```
fpichler: please do not use String representations for document full names. We are working about 10 years to get rid of all String typed full names. Use DocumentReferences or EntityReferences instead.

src/main/java/com/celements/model/classes/fields/AbstractClassField.java
msladek: I guess the naming is wrong since it does not represent the className, even though it actually is. It is only used to load the ClassDefinition and there should rather be called _classDefinitionName_ or similar. WDYT?

PR #73, opened 2016-09-03
Add Layout-Type to PageLayoutPropertiesClass

fpichler:
https://synjira.atlassian.net/browse/CELDEV-328
https://synjira.atlassian.net/browse/CELDEV-329
and lots of refactorings

### DIFF
--- a/src/main/java/com/celements/navigation/ListBuilder.java
+++ b/src/main/java/com/celements/navigation/ListBuilder.java
@@ -20,7 +20,8 @@
 package com.celements.navigation;
 
 import java.util.Arrays;
-import java.util.LinkedHashSet;
+import java.util.NavigableSet;
+import java.util.TreeSet;
 
 import org.apache.commons.lang.StringUtils;
 
@@ -37,18 +38,18 @@ public class ListBuilder implements INavigationBuilder {
   @Override
   @Deprecated
   public void openLevel(String mainUlCSSClasses) {
-    openLevel(new LinkedHashSet<String>(Arrays.asList(StringUtils.split(mainUlCSSClasses, ' '))));
+    openLevel(new TreeSet<String>(Arrays.asList(StringUtils.split(mainUlCSSClasses, ' '))));
   }
 
   @Override
-  public void openLevel(LinkedHashSet<String> mainUlCSSClasses) {
+  public void openLevel(NavigableSet<String> mainUlCSSClasses) {
     outStream.append("<ul");
     addCssClasses(mainUlCSSClasses);
     outStream.append(">");
     isFirstItem = true;
   }
 
-  private void addCssClasses(LinkedHashSet<String> cssClasses) {
+  private void addCssClasses(NavigableSet<String> cssClasses) {
     if (!cssClasses.isEmpty()) {
       outStream.append(" class=\"");
       outStream.append(StringUtils.join(cssClasses, ' '));
@@ -85,9 +86,9 @@ public class ListBuilder implements INavigationBuilder {
     return uniqueId;
   }
 
-  LinkedHashSet<String> getCssClasses(boolean isFirstItem, boolean isLastItem, boolean isActive,
-      LinkedHashSet<String> additionalCssClasses) {
-    LinkedHashSet<String> cssClass = new LinkedHashSet<>(additionalCssClasses);
+  NavigableSet<String> getCssClasses(boolean isFirstItem, boolean isLastItem, boolean isActive,
+      NavigableSet<String> additionalCssClasses) {
+    NavigableSet<String> cssClass = new TreeSet<>(additionalCssClasses);
     if (isFirstItem) {
       cssClass.add("first");
     }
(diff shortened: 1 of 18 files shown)


src/main/java/com/celements/navigation/ListBuilder.java
```diff
@@ -29,33 +36,47 @@ public ListBuilder(String navUniqueId) {
     this.uniqueName = navUniqueId;
   }
 
+  @Override
+  @Deprecated
   public void openLevel(String mainUlCSSClasses) {
-    outStream.append("<ul" + addCssClasses(mainUlCSSClasses) + ">");
+    openLevel(new TreeSet<String>(Arrays.asList(StringUtils.split(mainUlCSSClasses, ' '))));
```
msladek: why not LinkedHashSet here? TreeSet sorts mainUlCSSClasses lexicographically and does change the behaviour of openLevel.

src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -57,6 +62,48 @@
...
+
+  private final ConcurrentMap<String, IPageTypeCategoryRole> typeNameToCategoryMap = new ConcurrentHashMap<>();
+
+  private ConcurrentMap<String, IPageTypeCategoryRole> getTypeNameToCategoryMap() {
+    if (typeNameToCategoryMap.isEmpty()) {
+      synchronized (typeNameToCategoryMap) {
+        if (typeNameToCategoryMap.isEmpty()) {
+          for (IPageTypeCategoryRole typeCategory : pageTypeCategoryList) {
+            IPageTypeCategoryRole beforeRegCat = typeNameToCategoryMap.putIfAbsent(
+                typeCategory.getTypeName(), typeCategory);
```
msladek: perhaps better use com.google.common.collect.ImmutableMap.Builder instead of using and returning a modifiable ConcurrentHashMap.

src/main/java/com/celements/web/plugin/cmd/PageLayoutCommand.java
```diff
@@ -127,16 +131,15 @@ public String createNew(SpaceReference layoutSpaceRef) {
     if (layoutSpaceRef != null) {
       if (!layoutExists(layoutSpaceRef)) {
         try {
-          XWikiDocument propXdoc = getContext().getWiki().getDocument(standardPropDocRef(
-              layoutSpaceRef), getContext());
+          XWikiDocument propXdoc = getModelAccess().getOrCreateDocument(standardPropDocRef(
+              layoutSpaceRef));
           BaseObject layoutPropObj = propXdoc.newXObject(getPageLayoutPropertiesClassRef(
               propXdoc.getDocumentReference().getWikiReference().getName()), getContext());
```
msladek: use `com.celements.model.access.DefaultModelAccessFacade.newXObject(XWikiDocument, DocumentReference)` instead

src/main/java/com/celements/web/plugin/cmd/PageLayoutCommand.java
```diff
@@ -156,15 +159,14 @@ public boolean deleteLayout(SpaceReference layoutSpaceRef) {
       spaceDocsQuery.bindValue("space", layoutSpaceRef.getName());
       for (String docName : spaceDocsQuery.<String>execute()) {
         DocumentReference docReference = getWebUtilsService().resolveDocumentReference(docName);
-        xwiki.deleteAllDocuments(xwiki.getDocument(docReference, getContext()), getContext());
+        xwiki.deleteAllDocuments(getModelAccess().getDocument(docReference), getContext());
```
msladek: use `com.celements.model.access.IModelAccessFacade.deleteDocument(DocumentReference, boolean)` instead

src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -57,6 +62,48 @@
...
+            if (beforeRegCat != null) {
+              LOGGER.warn("Page type category collision on category name '{}' the colliding"
+                  + " category '{}' is shadowed by ''.", typeCategory.getTypeName(),
```
msladek: third `{}` missing for `beforeRegCat.getClass()`

src/main/java/com/celements/web/plugin/cmd/PageLayoutCommand.java
```diff
@@ -446,6 +449,15 @@ public String getPrettyName(SpaceReference layoutSpaceRef) {
     return "";
   }
 
+  public String getLayoutType(SpaceReference layoutSpaceRef) {
+    BaseObject layoutPropertyObj = getLayoutPropertyObj(layoutSpaceRef);
+    if ((layoutPropertyObj != null) && (layoutPropertyObj.getStringValue(
+        ICellsClassConfig.LAYOUT_TYPE_FIELD) != null)) {
+      return layoutPropertyObj.getStringValue(ICellsClassConfig.LAYOUT_TYPE_FIELD);
```
msladek: afaik `com.xpn.xwiki.objects.BaseCollection.getStringValue(String)` never returns null

src/main/java/com/celements/pagetype/service/PageTypeService.java
fpichler: this is for private PageTypeService use only. No Api, no external access.

msladek: I don't see the need for a ConcurrentMap since there is no simultaneous read/write to the map. Assigning a volatile variable inside synchronized should be sufficient.
I'd still consider immutability for clarity reasons: the map is not expected to change outside the synchronized block, therefore fast failure upon modification is preferable imho.

msladek:
see reply, otherwise good to merge

src/main/java/com/celements/pagetype/service/PageTypeService.java
fpichler: ConcurrentMap is the better choice to ensure memory visibility compared to volatile.  The map could change if a Category-implementation decides to change the handled Category name. The Category names are not defined to be static. I am even not completely if we should consider the deprecated names in the map. Yet, I agree that we could return an unmodifiable map from the getter method, because it is a cache which must be rebuilded so far. I added resetTypeNameToCatCache to allow any Category implementation to reset the cache if needed. Done.

fpichler:
"see reply"... which reply?

msladek:
[this reply](https://github.com/celements/celements-core/pull/73#discussion_r77600620) :)

makes more sense to me now with the changes, thanks!

PR #76, opened 2016-09-08
ModelAccess strategy

msladek:
https://synjira.atlassian.net/browse/CELDEV-335

### DIFF
--- a/src/main/java/com/celements/model/access/ContextExecutor.java
+++ b/src/main/java/com/celements/model/access/ContextExecutor.java
@@ -1,7 +1,5 @@
 package com.celements.model.access;
 
-import javax.validation.constraints.NotNull;
-
 import org.xwiki.model.reference.WikiReference;
 
 import com.celements.model.context.ModelContext;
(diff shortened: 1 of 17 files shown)


src/main/java/com/celements/model/access/ContextExecutor.java
```diff
@@ -0,0 +1,47 @@
...
+
+  public ContextExecutor<T, E> inWiki(WikiReference wiki) {
+    wiki = References.cloneRef(wiki, WikiReference.class);
+    return this;
+  }
+
+  public T execute() throws E {
+    WikiReference currWiki = getContext().getWikiRef();
+    try {
+      if (wiki != null) {
```
fpichler: I think checkState(wiki != null) is more appropriate to force a proper usage.

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -132,138 +101,69 @@ private XWikiDocument getDocumentCloneInternal(DocumentReference docRef) {
...
-    return exists;
+    return exists(docRef, DEFAULT_LANG);
   }
 
   @Override
   public boolean exists(DocumentReference docRef, String lang) {
     boolean exists = false;
     if (docRef != null) {
-      exists = existsFromStore(docRef, lang);
+      lang = normalizeLang(lang);
```
fpichler, line 141: please inline normalizeLang(lang) it is only used once on line 142.

src/main/java/com/celements/model/access/DefaultXWikiDocumentCreator.java
```diff
@@ -20,10 +20,10 @@
   private ModelContext context;
 
   @Override
-  public XWikiDocument createWithoutDefaults(DocumentReference docRef) {
+  public XWikiDocument createWithoutDefaults(DocumentReference docRef, String lang) {
```
fpichler, line 23: please readd old api createWithoutDefaults(DocumentReference) and mark deprecated.

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
msladek: it is also used on line 214

msladek:
> how can we select the modelAccessStrategy in a other project?

is this a requirement as of now? we could make the component configurable when the demand comes up for that because of a project specific strategy implementation, until then I would stick with YAGNI.

please re-review

PR #84, opened 2016-10-01
implementing getFieldValue (CELDEV-353)

fpichler:
CELDEV-353 : https://synjira.atlassian.net/browse/CELDEV-353

### DIFF
--- a/src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
+++ b/src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
@@ -40,6 +40,7 @@ import com.celements.rights.access.exceptions.NoAccessRightsException;
 import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
+import com.google.common.base.Optional;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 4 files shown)


fpichler:
please review and merge if okay.

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -633,6 +634,38 @@ private BaseProperty getBaseProperty(BaseObject obj, String key) {
   }
 
   @Override
+  public <T> Optional<T> getFieldValue(XWikiDocument doc, ClassField<T> field) {
+    return Optional.fromNullable(getProperty(checkNotNull(doc), checkNotNull(field)));
```
msladek, line 638: `checkNotNull(doc)` already happens in _getXObjects_ Line 430

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -633,6 +634,38 @@ private BaseProperty getBaseProperty(BaseObject obj, String key) {
...
+  }
+
+  @Override
+  public <T> Optional<T> getFieldValue(DocumentReference docRef, ClassField<T> field)
+      throws DocumentNotExistsException {
+    return Optional.fromNullable(getProperty(checkNotNull(docRef), checkNotNull(field)));
```
msladek, line 644: `checkNotNull(docRef)` already happens in _getDocumentReadOnly_ Line 136

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -633,6 +634,38 @@ private BaseProperty getBaseProperty(BaseObject obj, String key) {
...
+    checkNotNull(ignoreValue);
+    Optional<T> property = getFieldValue(docRef, field);
+    if (property.isPresent() && Objects.equal(property.get(), ignoreValue)) {
+      property = Optional.absent();
+    }
+    return property;
+  }
+
+  @Override
   public <T> T getProperty(DocumentReference docRef, ClassField<T> field)
```
msladek, line 669: perhaps we want to discourage using this method in favour of Optional return values by marking it as deprecated?

msladek (changes requested):
thanks for these additions

fpichler:
please fix first https://synjira.atlassian.net/browse/CELDEV-355

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
fpichler: This is right. Over five more embedded method calls in the current implementation. Yet, fast failure means checking early.

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
fpichler: This is right. Over four more embedded method calls in the current implementation. Yet, fast failure means checking early.

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
fpichler: this could be a later discussion. Is not the focus of this issue.

msladek (approved):

PR #85, opened 2016-10-03
implement CustomMapping for BooleanField

fpichler:
adding CustomMapping to BooleanField


msladek (commented):
added unit tests, looks good now

PR #86, opened 2016-10-05
ImmutableDocumentMetaData spaceRef/name

msladek:


src/main/java/com/celements/model/metadata/ImmutableDocumentMetaData.java
```diff
@@ -8,26 +8,31 @@
...
 import com.celements.model.util.References;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 
 @Immutable
 public final class ImmutableDocumentMetaData implements DocumentMetaData {
 
-  private final DocumentReference docRef;
+  private final SpaceReference spaceRef;
+  private final String name;
```
fpichler: suggestion: rename to docName

fpichler (changes requested):
just a little naming

PR #93, opened 2016-12-26
XObjectHandler

msladek:
XObjectHandler to fetch/create/remove XObjects on a XWikiDocument

https://synjira.atlassian.net/browse/CELDEV-299


src/main/java/com/celements/model/access/object/DefaultXObjectHandler.java
```diff
@@ -0,0 +1,247 @@
...
+
+  // do not use for XObject operations, could lead to endless loops
+  @Requirement
+  private IModelAccessFacade modelAccess;
+
+  @Requirement
+  private ModelContext context;
+
+  private XWikiDocument doc;
+  private final Map<DocumentReference, Map<String, Set<Object>>> filterMap = new LinkedHashMap<>();
```
fpichler, line 59: add some javadoc about what is expected to be in the filterMap. ClassDocRefs ?? and Map of?? or ??

src/main/java/com/celements/model/access/object/DefaultXObjectHandler.java
```diff
@@ -0,0 +1,247 @@
...
+      return ImmutableList.copyOf(getDoc().getXObjects().keySet());
+    } else {
+      final WikiReference wikiRef = References.extractRef(getDoc().getDocumentReference(),
+          WikiReference.class).get();
+      return FluentIterable.from(filterMap.keySet()).transform(new AdjustWikiFunction<>(
+          DocumentReference.class, wikiRef)).toList();
+    }
+  }
+
+  private Map<String, Set<Object>> getValueMap(DocumentReference classRef, boolean add) {
```
fpichler, line 85: refactor in three methods getValueMap (could return Optional), containsClassRefInFilterMap and addClassRefToFilterMap. Idea: FilterMap could be a private inner class. Add javadoc that these methods expect normalized classRefs instead of normalizing the class refs in the internal methods again and again

src/main/java/com/celements/model/access/object/DefaultXObjectHandler.java
```diff
@@ -0,0 +1,247 @@
...
+    return filter(field, Arrays.asList(value));
+  }
+
+  @Override
+  public <T> XObjectHandler filter(ClassField<T> field, Collection<T> values) {
+    return filter(checkNotNull(field).getClassDef().getClassRef(), field.getName(), values);
+  }
+
+  @Override
+  public XObjectHandler filter(DocumentReference classRef, String key, Object value) {
```
fpichler, line 117: please add deprecated annotation to the implementation too

src/main/java/com/celements/model/access/object/DefaultXObjectHandler.java
```diff
@@ -0,0 +1,247 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.xpn.xwiki.XWikiException;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.objects.BaseObject;
+
+@NotThreadSafe
+@Component
+@InstantiationStrategy(ComponentInstantiationStrategy.PER_LOOKUP)
+public class DefaultXObjectHandler implements XObjectHandler {
```
fpichler, line 47: split class in a XWikiXObjectHandler (do not deprecate but add a hint that better type safty can be achieved by using ClassFieldXObjectHandler) with Strings and Objects for key and values and a ClassFieldXObjectHandler only using ClassFields<T> and values T too in the FilterMap(s)

fpichler (changes requested):
A very nice idea for a core improvement. Maybe add a factory method to ModelAccess to get a new XWikiXObjectHandler or a new ClassFieldXObjectHandler object easily. Overall it seams not quite finished to me. The internal FilterMap is complicated to understand, possibly fragile because of using Strings and Object types and could possibly draw more out of the ClassDefinitions and ClassField core feature.

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -275,105 +272,90 @@ private String normalizeLang(String lang) {
...
     return Iterables.getFirst(getXObjects(getDocumentReadOnly(docRef, DEFAULT_LANG), classRef, key,
         value), null);
   }
 
   @Override
+  @Deprecated
   public BaseObject getXObject(XWikiDocument doc, DocumentReference classRef) {
-    return Iterables.getFirst(getXObjects(doc, classRef), null);
+    return Utils.getComponent(XObjectHandler.class).onDoc(doc).filter(
+        classRef).fetchFirst().orNull();
```
ebeutler, line 293: Why isn't this a @Requirement? ... I assume cyclical dependancy. If so could you add a comment, so we remember in 3 months? (maybe extract method, since it gets used quite often)

msladek:
continued on branch CELDEV-299

PR #95, opened 2017-01-08
Newfeature meta tag collector

ebeutler:
https://synjira.atlassian.net/browse/CELDEV-390


src/main/java/com/celements/metatag/IMetaTag.java
```diff
@@ -0,0 +1,14 @@
+package com.celements.metatag;
+
+import org.xwiki.component.annotation.ComponentRole;
+
+@ComponentRole
+public interface IMetaTag {
```
msladek: we once decided to drop the I for interface naming and rather name the implementations according to what they actually are doing, e.g. `StoreModelAccessStrategy implements ModelAccessStrategy`

also, isn't this rather a MetaTagService?

src/main/java/com/celements/metatag/MetaTag.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  @Override
+  public String displayAllMetaTags() {
+    StringBuilder sb = new StringBuilder();
+    Object contextObj = modelContext.getXWikiContext().remove(IMetaTag.META_CONTEXT_KEY);
+    if (contextObj instanceof List) {
+      @SuppressWarnings("unchecked")
+      List<MetaTagApi> metaList = (List<MetaTagApi>) contextObj;
```
msladek: DRY, extract as e.g. `public List<MetaTagApi> getMetaTags()`

src/main/java/com/celements/metatag/MetaTagApi.java
```diff
@@ -0,0 +1,94 @@
...
+import com.celements.metatag.enums.EReferrer;
+import com.celements.metatag.enums.ERobot;
+import com.celements.metatag.enums.EViewport;
+import com.celements.metatag.enums.opengraph.EOpenGraph;
+import com.celements.metatag.enums.twitter.ETwitter;
+import com.celements.metatag.enums.twitter.ETwitterCardType;
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableMap;
+
+public final class MetaTagApi {
```
msladek: with the MetaTagService, this can now be called MetaTag

msladek (changes requested):

src/main/java/com/celements/metatag/IMetaTag.java
fpichler: :+1:  for MetaTagServiceRole

src/main/java/com/celements/metatag/MetaTag.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Component
+public class MetaTag implements IMetaTag {
+
+  @Requirement
+  ModelContext modelContext;
+
+  @Override
+  public void addMetaTag(MetaTagApi tag) {
+    Object contextObj = modelContext.getXWikiContext().get(IMetaTag.META_CONTEXT_KEY);
```
fpichler: Better use Execution context instead of deprecated XWikiContext to store/cache the metaList for one request

src/main/java/com/celements/metatag/MetaTag.java
```diff
@@ -0,0 +1,45 @@
...
+    if (contextObj instanceof List) {
+      @SuppressWarnings("unchecked")
+      List<MetaTagApi> metaList = (List<MetaTagApi>) contextObj;
+      metaList.add(tag);
+    } else {
+      List<MetaTagApi> metaList = new ArrayList<>();
+      metaList.add(tag);
+      modelContext.getXWikiContext().put(IMetaTag.META_CONTEXT_KEY, metaList);
```
fpichler: Better use Execution context instead of deprecated XWikiContext to store/cache the metaList for one request

src/main/java/com/celements/metatag/MetaTag.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import org.xwiki.component.annotation.Component;
+import org.xwiki.component.annotation.Requirement;
+
+import com.celements.model.context.ModelContext;
+
+@Component
+public class MetaTag implements IMetaTag {
```
fpichler: I think, this is then the MetaTagService or DefaultMetaTagService ...

src/main/java/com/celements/metatag/MetaTagApi.java
fpichler: :-1:  MetaTagService
xxxApi are objects which might be used only from scripts (velocity, groovy, ...) The xxxApi is always a wrapper/facade concerned to prevent rights escalations.

It could be that this should be a ScriptService implementation. This would mean that the skripting way to access this singleton is services.xxx . In contrast to the MetaTagApi object being returned by a different ScriptService call.

fpichler (changes requested):
please send back to review after changes are made

src/main/java/com/celements/metatag/MetaTagApi.java
msladek: Like you said, I mostly understood the XXXApi class to be a wrapper which prevents rights escalation for an XXX class. The latter doesn't exist in this case though.

I know that the "Api"-naming stems from XWiki, but I would actually prefer a discrete annotation to generally mark java classes which are passed to scripting languages, e.g.:

`@Scripting
public final class MetaTag`

E.g. we don't put "Singleton" into the naming since we can note it `@Singleton`.

fpichler: Interressting idea but not discussed and decided upon. PR is the wrong please to bring up new codeing style ideas. Please take this discussion offline of this PR and stick to the style guide /naming convention in use.

ebeutler:
Changed naming and now uses ExecutionContext to collect meta tags

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -141,7 +141,7 @@
   LastStartupTimeStampRole lastStartupTimeStamp;
 
   @Requirement
-  IMetaTag metaTag;
+  MetaTagSerivceRole metaTag;
```
fpichler: please rename MetaTagSerivceRole to MetaTagServiceRole (spelling!!)

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1077,25 +1077,26 @@ public String getVersion(String systemComponentName) {
     return this.versionMap.get(systemComponentName);
   }
 
-  public MetaTagApi getCharsetMetaTag(String charsetStr) {
+  public MetaTag getCharsetMetaTag(String charsetStr) {
```
fpichler: please move MetaTag ScriptService calls to own ScriptService class e.g. MetaTagScriptService ("metatag").

fpichler (changes requested):

fpichler (approved):

src/main/java/com/celements/metatag/MetaTag.java
```diff
@@ -0,0 +1,94 @@
...
+
+  public MetaTag(@NotNull ETwitter twitter, @NotNull String content) {
+    this(ETwitter.ATTRIB_NAME, twitter.getIdentifier(), content);
+  }
+
+  public MetaTag(@NotNull ETwitterCardType twitterCardType) {
+    this(ETwitter.TWITTER_CARD, twitterCardType.getIdentifier());
+  }
+
+  public String display() {
```
msladek: NotNull/Nullable missing

src/main/java/com/celements/metatag/MetaTagScriptService.java
```diff
@@ -0,0 +1,41 @@
...
+
+@Component(MetaTagScriptService.COMPONENT_HINT)
+public class MetaTagScriptService {
+
+  public static final String COMPONENT_HINT = "metatag";
+
+  @Requirement
+  MetaTagServiceRole metaTag;
+
+  public MetaTag getCharsetMetaTag(String charsetStr) {
```
msladek: NotNull/Nullable missing

src/main/java/com/celements/metatag/MetaTagScriptService.java
```diff
@@ -0,0 +1,41 @@
...
+      @NotNull String content) {
+    metaTag.addMetaTagToCollector(new MetaTag(attributeName, attributeValue, content));
+  }
+
+  public void addMetaTagToCollector(@NotNull Map<String, String> attributes,
+      @NotNull String content) {
+    metaTag.addMetaTagToCollector(new MetaTag(attributes, content));
+  }
+
+  public String displayCollectedMetaTags() {
```
msladek: NotNull/Nullable missing

src/main/java/com/celements/metatag/MetaTagServiceRole.java
```diff
@@ -0,0 +1,14 @@
+package com.celements.metatag;
+
+import org.xwiki.component.annotation.ComponentRole;
+
+@ComponentRole
+public interface MetaTagServiceRole {
+
+  public static final String META_CONTEXT_KEY = "celements_meta_tags";
+
+  public void addMetaTagToCollector(MetaTag tag);
```
msladek: NotNull/Nullable missing

src/main/java/com/celements/metatag/MetaTagServiceRole.java
```diff
@@ -0,0 +1,14 @@
...
+
+  public String displayCollectedMetaTags();
```
msladek: NotNull/Nullable missing

msladek (changes requested):

msladek:
a few NotNull/Nullable missing but looks all good to merge apart from that

msladek (approved):

PR #96, opened 2017-01-11
checkNotTranslation error message fix

msladek:
https://synjira.atlassian.net/browse/CELDEV-408

### DIFF
--- a/src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
+++ b/src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
@@ -2,6 +2,7 @@ package com.celements.model.access;
 
 import static com.google.common.base.Preconditions.*;
 
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -359,9 +358,10 @@ public BaseObject getXObject(XWikiDocument doc, DocumentReference classRef, Stri
   }
 
   private void checkNotTranslation(XWikiDocument doc) {
-    String msg = "Trying to access XObjects on translation '{0}' of doc '{1}";
-    checkState(!isTranslation(doc), MessageFormat.format(msg, doc.getLanguage(),
-        doc.getDocumentReference()));
+    if (isTranslation(doc)) {
```
fpichler: keep using MessageFormat.format but fix the wrong usage. I expect that the single quotes cause the problem. Extract the Error-Message to a constant and check in a simple unit test that the replacement with MessageFormat is working.

fpichler (changes requested):

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -358,8 +358,8 @@ public BaseObject getXObject(XWikiDocument doc, DocumentReference classRef, Stri
     return Collections.unmodifiableMap(ret);
   }
 
-  private void checkNotTranslation(XWikiDocument doc) {
-    String msg = "Trying to access XObjects on translation '{0}' of doc '{1}";
+  void checkNotTranslation(XWikiDocument doc) {
+    String msg = "Trying to access XObjects on translation ''{0}'' of doc ''{1}''";
```
fpichler: please extract to constant.

fpichler (changes requested):

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
msladek: what for? it is only used in this exact method.

fpichler: for better readability. You may inline instead, but do not assign a constant final string to a local variable.

fpichler (approved):

PR #97, opened 2017-01-11
DocFormCommand NPE for no template

msladek:
https://synjira.atlassian.net/browse/CELDEV-410

### DIFF
--- a/src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
+++ b/src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
@@ -2,6 +2,7 @@ package com.celements.model.access;
 
 import static com.google.common.base.Preconditions.*;
 
+import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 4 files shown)


src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -359,9 +358,10 @@ public BaseObject getXObject(XWikiDocument doc, DocumentReference classRef, Stri
   }
 
   private void checkNotTranslation(XWikiDocument doc) {
-    String msg = "Trying to access XObjects on translation '{0}' of doc '{1}";
-    checkState(!isTranslation(doc), MessageFormat.format(msg, doc.getLanguage(),
-        doc.getDocumentReference()));
+    if (isTranslation(doc)) {
```
fpichler: keep using MessageFormat.format but fix the wrong usage. I expect that the single quotes cause the problem. Extract the Error-Message to a constant and check in a simple unit test that the replacement with MessageFormat is working.

src/main/java/com/celements/web/plugin/cmd/DocFormCommand.java
```diff
@@ -319,24 +321,28 @@ String getFieldFromProperty(PropertyClass propclass, String field) {
 
   private BaseClass getBaseClass(String className, XWikiContext context) {
```
fpichler: drop not used XWikiContext parameter (signature change only because it is a private method)

fpichler (changes requested):

src/main/java/com/celements/model/access/DefaultModelAccessFacade.java
```diff
@@ -357,11 +358,10 @@ public BaseObject getXObject(XWikiDocument doc, DocumentReference classRef, Stri
     return Collections.unmodifiableMap(ret);
   }
 
-  private void checkNotTranslation(XWikiDocument doc) {
-    if (isTranslation(doc)) {
-      throw new IllegalStateException("Trying to access XObjects on translation '"
-          + doc.getLanguage() + "' of doc '" + doc.getDocumentReference() + "'");
-    }
+  void checkNotTranslation(XWikiDocument doc) {
+    String msg = "Trying to access XObjects on translation ''{0}'' of doc ''{1}''";
```
fpichler: please extract to constant

fpichler (changes requested):

PR #98, opened 2017-01-17
convert page type properties class config to class definition

gcohen08:
https://synjira.atlassian.net/browse/CELDEV-393

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 /.settings
 /xwiki.log*
 /src/main/java/META-INF
+/bin/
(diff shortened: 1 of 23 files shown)


fpichler (changes requested):
If the PageTypeClasses.java are refactored with a new ClassDefinition, then please remove the class generation from the initClasses method in PageTypeClasses. Otherwise the same class gets created in two different places. Furthermore please mark the PageTypeClasses class as deprecated with a javadoc "instead use..." hint.

msladek (approved):

msladek:
thanks a lot for the refactoring

PR #100, opened 2017-01-31
fix bug not getting page types for empty category name

ebeutler:
https://synjira.atlassian.net/browse/CELDEV-406

### DIFF
--- a/src/main/java/com/celements/pagetype/service/PageTypeService.java
+++ b/src/main/java/com/celements/pagetype/service/PageTypeService.java
@@ -66,13 +66,15 @@ public class PageTypeService implements IPageTypeRole {
   private List<IPageTypeCategoryRole> pageTypeCategoryList;
 
   private final ConcurrentMap<String, IPageTypeCategoryRole> typeNameToCatCache = new ConcurrentHashMap<>();
+  private final ConcurrentMap<String, IPageTypeCategoryRole> typeNameToCatCacheDeprecated = new ConcurrentHashMap<>();
 
   @Override
   public void resetTypeNameToCatCache() {
     typeNameToCatCache.clear();
+    typeNameToCatCacheDeprecated.clear();
   }
 
-  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMap() {
+  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMap(boolean deprecated) {
     if (typeNameToCatCache.isEmpty()) {
       synchronized (typeNameToCatCache) {
         if (typeNameToCatCache.isEmpty()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -66,13 +66,15 @@
...
+  private final ConcurrentMap<String, IPageTypeCategoryRole> typeNameToCatCacheDeprecated = new ConcurrentHashMap<>();
 
   @Override
   public void resetTypeNameToCatCache() {
     typeNameToCatCache.clear();
+    typeNameToCatCacheDeprecated.clear();
   }
 
-  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMap() {
+  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMap(boolean deprecated) {
```
fpichler: please do not add a boolean to the getTypeNameToCategoryMap. This binds the implementation to the interface. Instead rename getTypeNameToCategoryMap to void loadTypeNameToCategoryMap. Then implement getTypeNameToCategoryMap returning only the not deprecated names (as before) and getTypeNameToCategoryMapIncludeDeprecated returning typeNameToCatCacheDeprecated. Call the loadTypeNameToCategoryMap at the beginning of both get methods.

fpichler (changes requested):

src/main/java/com/celements/pagetype/service/PageTypeService.java
fpichler: More Details in Martin Fowler: Refactoring p. 279 (Separate Query from Modifier)

ebeutler:
changed, even though, I'm not sure why changing the signature of a private method should bind the implementation to the interface.

src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -84,16 +86,36 @@ public void resetTypeNameToCatCache() {
                   + " category '{}' is shadowed by '{}'.", typeCategory.getTypeName(),
                   typeCategory.getClass(), beforeRegCat.getClass());
             }
+            for (String deprecateName : typeCategory.getDeprecatedNames()) {
```
fpichler: use getAllTypeNames instead of getDeprecatedNames

src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -84,16 +86,36 @@ public void resetTypeNameToCatCache() {
...
+
+  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMap() {
+    loadTypeNameToCategoryMap();
     return Collections.unmodifiableMap(typeNameToCatCache);
   }
 
+  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMapIncludeDeprecated() {
+    loadTypeNameToCategoryMap();
+    ConcurrentMap<String, IPageTypeCategoryRole> completeMap = typeNameToCatCacheDeprecated;
+    completeMap.putAll(typeNameToCatCache);
```
fpichler: using use getAllTypeNames instead of getDeprecatedNames: this readding over and over again is not needed.

src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -84,16 +86,36 @@ public void resetTypeNameToCatCache() {
...
+    return Collections.unmodifiableMap(completeMap);
+  }
+
   @Override
   public Optional<IPageTypeCategoryRole> getTypeCategoryForCatName(String categoryName) {
-    return Optional.fromNullable(getTypeNameToCategoryMap().get(categoryName));
+    loadTypeNameToCategoryMap();
```
fpichler: additional loadTypeNameToCategoryMap not needed.

src/main/java/com/celements/pagetype/service/PageTypeService.java
```diff
@@ -66,13 +66,15 @@
...
+  private final ConcurrentMap<String, IPageTypeCategoryRole> typeNameToCatCacheDeprecated = new ConcurrentHashMap<>();
 
   @Override
   public void resetTypeNameToCatCache() {
     typeNameToCatCache.clear();
+    typeNameToCatCacheDeprecated.clear();
   }
 
-  private Map<String, IPageTypeCategoryRole> getTypeNameToCategoryMap() {
+  private void loadTypeNameToCategoryMap() {
```
fpichler: maybe plural is better loadTypeNameCategoryMaps because we load to different maps.

fpichler (changes requested):

fpichler (approved):

PR #104, opened 2017-02-22
add component hook to add meta tags to the header

ebeutler:
https://synjira.atlassian.net/browse/CELDEV-389


src/main/java/com/celements/metatag/MetaTagHeaderRole.java
```diff
@@ -0,0 +1,13 @@
+package com.celements.metatag;
+
+import java.util.List;
+
+import javax.validation.constraints.NotNull;
+
+import org.xwiki.component.annotation.ComponentRole;
+
+@ComponentRole
+public interface MetaTagHeaderRole {
```
msladek: rename to MetaTagProviderRole and add `public @NotNull List<MetaTag> getBodyMetaTags();` as well since meta tags are allowed in body too for HTML5

src/main/java/com/celements/metatag/MetaTagServiceRole.java
```diff
@@ -13,4 +13,6 @@
 
   public @NotNull String displayCollectedMetaTags();
 
+  public void loadHeaderTags();
```
msladek: collectHeaderTags / collectBodyTags

msladek (changes requested):

msladek (approved):

fpichler (approved):

PR #106, opened 2017-02-28
Tests are all green

nderguti:


src/main/java/com/celements/metatag/enums/twitter/ETwitterCardType.java
```diff
@@ -30,8 +29,8 @@ public String getIdentifier() {
   }
 
   @NotNull
-  public static Optional<ETwitterCardType> getTwitterCardType(@Nullable String identifier) {
-    return ID_MAP.get(identifier);
+  public static ETwitterCardType getTwitterCardType(@Nullable String identifier) {
```
msladek: keep Optional return parameter in all enums

msladek (changes requested):

src/main/java/com/celements/metatag/enums/twitter/ETwitterCardType.java
nderguti: @msladek I will keep it in mind to check better, thank you that you did it, I missed this one

PR #107, opened 2017-03-31
isLoggedIn() method 

nderguti:
https://synjira.atlassian.net/browse/CELDEV-472

### DIFF
--- a/src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
+++ b/src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
@@ -82,13 +82,12 @@ public class DefaultRightsAccessFacade implements IRightsAccessFacadeRole {
 
   @Override
   public boolean isLoggedIn() {
-    boolean ret;
+    boolean ret = false;
     if (!"XWiki.XWikiGuest".equals(getContext().getUser())) {
-      LOGGER.info("User is logged in");
       ret = true;
+      LOGGER.info("User is logged in", ret);
     } else {
-      ret = false;
-      LOGGER.info("User is not logged in");
+      LOGGER.info("User is not logged in", ret);
     }
     return ret;
   }
--- a/src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
+++ b/src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
@@ -79,4 +79,13 @@ public class DefaultRightsAccessFacade implements IRightsAccessFacadeRole {
         user, ret);
     return ret;
   }
+
+  public boolean isLoggedIn() {
+    if ("XWiki.XWikiGuest".equals(getContext().getUser())) {
+      return true;
+    } else {
+      LOGGER.info("User is not logged in");
+    }
+    return false;
+  }
 }
--- a/src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
+++ b/src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
@@ -81,7 +81,7 @@ public class DefaultRightsAccessFacade implements IRightsAccessFacadeRole {
   }
 
   public boolean isLoggedIn() {
-    if ("XWiki.XWikiGuest".equals(getContext().getUser())) {
+    if (!"XWiki.XWikiGuest".equals(getContext().getUser())) {
       return true;
     } else {
       LOGGER.info("User is not logged in");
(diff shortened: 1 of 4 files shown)


src/main/java/com/celements/rights/access/RightsAccessScriptService.java
```diff
@@ -56,4 +60,10 @@ public boolean hasAccessLevel(DocumentReference docRef, EAccessLevel level, XWik
     return hasAccessLevel(docRef, level, user);
   }
 
+  public boolean isLoggedIn() {
+    if (!"XWiki.XWikiGuest".equals(xWikiContext.getUser())) {
+      return true;
+    }
+    return false;
```
msladek: use `rightsAccess.isLoggedIn()` instead of reimplementing the method

src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
```diff
@@ -79,4 +79,13 @@ public boolean hasAccessLevel(EntityReference ref, EAccessLevel level, XWikiUser
         user, ret);
     return ret;
   }
+
+  public boolean isLoggedIn() {
+    if (!"XWiki.XWikiGuest".equals(getContext().getUser())) {
+      return true;
```
msladek: for clarity reasons, it's preferable to only have one return statement at the end of the method, e.g.:
```
boolean ret = ...;
LOGGER...
return ret;

**Review by msladek — CHANGES_REQUESTED** [review]

**Comment by msladek** [comment]
also add unit tests for isLoggedIn in DefaultRightsAccessFacadeTest

File: src/main/java/com/celements/rights/access/RightsAccessScriptService.java [hunk 3]
```diff
@@ -22,6 +23,9 @@
   @Requirement
   IRightsAccessFacadeRole rightsAccess;
 
+  @Requirement
+  XWikiContext xWikiContext;
```
ebeutler: 1. XWikiContext isn't a component, so this wouldn't work.
2. xWikiContext is not used.
-> please remove

src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
```diff
@@ -79,4 +79,17 @@ public boolean hasAccessLevel(EntityReference ref, EAccessLevel level, XWikiUser
         user, ret);
     return ret;
   }
+
+  @Override
+  public boolean isLoggedIn() {
+    boolean ret;
+    if (!"XWiki.XWikiGuest".equals(getContext().getUser())) {
```
ebeutler: Please use XWikiRightService.GUEST_USER_FULLNAME instead of hard coded String (same applies in the tests)

src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
```diff
@@ -79,4 +79,17 @@ public boolean hasAccessLevel(EntityReference ref, EAccessLevel level, XWikiUser
         user, ret);
     return ret;
   }
+
+  @Override
+  public boolean isLoggedIn() {
+    boolean ret;
```
ebeutler: Please initialise to make the current value explicit.

src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
```diff
@@ -79,4 +79,17 @@ public boolean hasAccessLevel(EntityReference ref, EAccessLevel level, XWikiUser
...
+    if (!"XWiki.XWikiGuest".equals(getContext().getUser())) {
+      LOGGER.info("User is logged in");
+      ret = true;
+    } else {
+      ret = false;
```
ebeutler: No need to set again, since this is the same as the default value.

ebeutler (changes requested):

ebeutler (approved):

src/main/java/com/celements/rights/access/DefaultRightsAccessFacade.java
```diff
@@ -79,4 +79,16 @@ public boolean hasAccessLevel(EntityReference ref, EAccessLevel level, XWikiUser
         user, ret);
     return ret;
   }
+
+  @Override
+  public boolean isLoggedIn() {
+    boolean ret = false;
+    if (!XWikiRightService.GUEST_USER_FULLNAME.equals(getContext().getUser())) {
```
msladek: again there is no need for an if. also, the logger never prints `ret`. why not do it like that?
```
boolean ret = ...;
LOGGER.info("isLoggedIn: {}", ret);
return ret;
```

msladek (changes requested):

msladek (approved):

PR #112, opened 2017-06-27
ClassDefinition CustomListField & Marshallers

msladek:
https://synjira.atlassian.net/browse/CELDEV-510


src/main/java/com/celements/model/classes/fields/list/ListField.java
```diff
@@ -82,21 +92,26 @@ protected ListField(@NotNull Builder<?, T> builder) {
...
+    List<?> list = Collections.emptyList();
     if (obj instanceof String) {
       list = Splitter.on(getSeparator()).splitToList((String) obj);
-    } else if (obj != null) {
+    } else if (obj instanceof List) {
       list = getType().cast(obj);
-    } else {
-      list = new ArrayList<>();
+    } else if (obj != null) {
+      // TODO log
```
ebeutler: open TODO

ebeutler (changes requested):

src/main/java/com/celements/marshalling/Marshaller.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.google.common.base.Function;
+import com.google.common.base.Optional;
+
+@Immutable
+public interface Marshaller<T> {
+
+  public @NotNull Class<T> getToken();
+
+  public @NotNull Object serialize(@NotNull T val);
```
fpichler: Replace Object with String. If a more general solution is needed (use case ?) then use generic S instead.

fpichler (changes requested):

fpichler (approved):

ebeutler (approved):

PR #115, opened 2017-07-12
AbstractClassPackage must support activated_classcollections

msladek:
https://synjira.atlassian.net/browse/CELDEV-516


fpichler (changes requested):
see Issue

fpichler (approved):

PR #117, opened 2017-07-21
Use ImmutableDocumentReference in DocumentBuilder

msladek:
https://synjira.atlassian.net/browse/CELDEV-522

first review #116


ebeutler (approved):

fpichler (approved):

msladek:
while testing i noticed that XWikiHibernateStore mutates the DocumentReference, I therefore made it mutable before calling the Store and immutable afterwards.

please check the changes again

fpichler (changes requested):
I would prefer to replace the XWikiHibernateStore class with our own fixed version which enforces the immutable references (must be a subclass of XWikiHibernateStore and it must overwrite the default component). I plan to do performance improvements later on on this overwritten class. Wdyt?

msladek:
yes, I like the idea.

msladek:
please review

src/main/java/com/celements/store/CelHibernateStore.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public void saveXWikiDoc(XWikiDocument doc, final XWikiContext context,
+      final boolean bTransaction) throws XWikiException {
+    // XWikiHibernateStore.saveXWikiDoc requires a mutable docRef
+    DocRefMutabilityExecutor<Void> exec = new DocRefMutabilityExecutor<Void>() {
+
+      @Override
+      public Void call(XWikiDocument doc) throws XWikiException {
+        CelHibernateStore.super.saveXWikiDoc(doc, context, bTransaction);
```
fpichler, line 29: Why not just copy the saveXWikiDoc implementation and adjust the setWikiReference line with a doc.setDocumentReference(References.adjustRef(... ?

src/main/java/com/celements/store/CelHibernateStore.java
```diff
@@ -0,0 +1,101 @@
...
+
+  @Override
+  public XWikiDocument loadXWikiDoc(XWikiDocument doc, final XWikiContext context)
+      throws XWikiException {
+    // XWikiHibernateStore.loadXWikiDoc requires a mutable docRef
+    DocRefMutabilityExecutor<XWikiDocument> exec = new DocRefMutabilityExecutor<XWikiDocument>() {
+
+      @Override
+      public XWikiDocument call(XWikiDocument doc) throws XWikiException {
+        return CelHibernateStore.super.loadXWikiDoc(doc, context);
```
fpichler, line 46: why not just copy the loadXWikiDoc implementation and adjust the doc.setDatabase(context.getDatabase()); line with a doc.setDocumentReference(References.adjustRef(... ?

fpichler (changes requested):
Please overwite and fix the implementations in CelHibernateStore for saveXWikiDoc and loadXWikiDoc to correctly work with immutable references

ebeutler (dismissed):

src/main/java/com/celements/store/CelHibernateStore.java
msladek: and what about the docRefs contained in objects? would those be immutable as well after loadXWikiDoc/saveXWikiDoc? it is not obvious to me since I find those methods extremely hard to understand. I think its not the scope of this PR, so I suggest the follow up issues:
1. copy loadXWikiDoc/saveXWikiDoc from XWikiHibernateStore and refactor to achieve
   - improved readability
   - immutable references so make DocRefMutabilityExecutor obsolete
1. improve performance by reducing amount of hqls executed

fpichler: The docRefs contained in objects are created over BaseObject.setName and BaseObject.setClassName. To solve this problem we would need to replace the DocumentReferenceResolver implementations with implementations returning ImmutableDocumentReferences.

Nevertheless please create the two issues for inlining the methods and improving load performance afterwards. Moreover I think on line 750-754 objects could get lost on loading! Please mention on the issue. Mention the issue numbers in CelHibernateStore. I will then approve :-)

msladek: added issues and mentioned in CelHibernateStore
https://synjira.atlassian.net/browse/CELDEV-530
https://synjira.atlassian.net/browse/CELDEV-531

fpichler (dismissed):

fpichler (approved):

PR #121, opened 2017-07-31
Model ObjectFetcher/Editor

msladek:
https://synjira.atlassian.net/browse/CELDEV-299


fpichler (changes requested):
Very nice work. :+1:  I really like the way you designed the Fetcher idea. YET, I was wondering if we could replace the ClassReferences with ClassDefinition instead. This would e.g. allow to double check on the ClassField given in getObjectField and much more details (maybe default values on creating a new object) could be taken from the ClassDefinition. Moreover the ClassDefinition can deliver the ClassReferences if needed.

msladek:
Thanks for the feedback! I would not entirely replace the ClassReference with ClassDefinition:
1. I rely on equals/hashCode, which are not implemented for ClassDefintions
2. It would make the ObjectFetcher/Editor exclusively usable for new classes or refactored ClassCollections to ClassDefinitions, ClassReference is more general
3. We can still use ClassDefinition functionality if available, a ClassDefinition can always be looked up since its component name is defined as the serialized ClassReference `CLASS_DEF_HINT = SPACE_NAME + "." + DOC_NAME` or via `ClassField.getClassDef()` in e.g. _getObjectField_

providing a `filter(ClassDefinition)` is a good idea though.

WDYT?

fpichler:
ok, I agree. We better implement a different second version of ObjectFetcher/Editor later on.
Yet, could you try to replace the effectiv classes in the signatures with interfaces as much as possible? E.g. Replacing BaseObject with ObjectInterface, maybe introducing an Interface for ClassReference, and replacing XWikiDocument with DocumentModelBridge...

msladek:
I've now introduced `interface ClassIdentity`, which is implemented by `ClassReference` and `ClassDefiniton`. Please check the changes if I understood correctly what you had in mind.

Sadly, `DocumentModelBridge` and `ObjectInterface` are not very well designed or useful interfaces, they lack crucial methods like get/create/deleteXObjects and getXClassReference/getNumber respectively, so I cannot use them in XWikiFetcher/-Editor/-Bridge.
That's also why I found it more helpful to define them completely generic as _D_ and _O_ in the interfaces, but a `ObjectBridge<D, O>` has to be provided by implementations. With this setup we are completely free to use XWiki-Classes atm as well as our own in the future. Our CelementsObject does not have to implement ObjectInterface.

fpichler:
Actually, I did not think about handling ClassDefinition and ClassReference using an interface, but I like your idea. :+1:  I was much more hinting to "Effective Java, Item 25: Design method signature carefully -> For parameter types, favor interfaces over classes". I think it would be much more flexible for future changes if we could use more interfaces.

src/main/java/com/celements/model/access/object/restriction/ObjectQuery.java
```diff
@@ -0,0 +1,75 @@
...
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+import com.celements.model.classes.ClassIdentity;
+import com.google.common.base.Function;
+import com.google.common.base.Predicate;
+import com.google.common.collect.FluentIterable;
+
+@NotThreadSafe
+public class ObjectQuery<O> extends LinkedHashSet<ObjectRestriction<O>> {
```
fpichler: Please do not extend the fundation Collection classes. XWikiContext is a very bad example in this. See "Item 14: Favour composition over inheritance" Is it possible to do here the same with composition instead of inheritance? ObjectQueryBuilder is only using add and addAll which would need to be proxied...

fpichler (changes requested):

fpichler (approved):

msladek:
@ebeutler please review

ebeutler (approved):

PR #132, opened 2017-09-04
get url from preferences and properties file

dkhasham:
related Ticket: https://synjira.atlassian.net/browse/SYNCEL-22


ebeutler (changes requested):
According to its name the method returns specifically the support link URL. So there is no need for a key (use a constant instead).

Alternatively you could give a more general name to the method.

ebeutler (approved):

PR #136, opened 2017-10-17
Send redirect and transform url to utf8

dkhasham:
https://synjira.atlassian.net/browse/PROGON-648

### DIFF
--- a/src/main/java/com/celements/web/service/CelementsWebScriptService.java
+++ b/src/main/java/com/celements/web/service/CelementsWebScriptService.java
@@ -1104,13 +1104,18 @@ public class CelementsWebScriptService implements ScriptService {
   }
 
   public String encodeUrlToUtf8(String urlStr) {
-    urlStr = urlStr.replaceAll("http://", "");
-    try {
-      urlStr = URLEncoder.encode(urlStr, "UTF-8");
-    } catch (UnsupportedEncodingException exp) {
-      LOGGER.error("Failed to encode url [" + urlStr + "] to utf-8", exp);
+    String pattern = "://";
+    int findIndex = urlStr.indexOf(pattern);
+    if (findIndex > 0) {
+      String protocol = urlStr.substring(0, findIndex + pattern.length());
+      String mainUrl = urlStr.substring(findIndex + pattern.length(), urlStr.length());
+      try {
+        urlStr = URLEncoder.encode(mainUrl, "UTF-8");
+      } catch (UnsupportedEncodingException exp) {
+        LOGGER.error("Failed to encode url [" + urlStr + "] to utf-8", exp);
+      }
+      urlStr = protocol + urlStr;
     }
-    urlStr = "http://" + urlStr;
     return urlStr;
   }
 
--- a/src/main/java/com/celements/web/service/CelementsWebScriptService.java
+++ b/src/main/java/com/celements/web/service/CelementsWebScriptService.java
@@ -1121,9 +1121,6 @@ public class CelementsWebScriptService implements ScriptService {
 
   public void sendRedirect(String urlStr) {
     try {
-      System.out.println(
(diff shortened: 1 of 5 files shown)


src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1100,4 +1103,24 @@ public String getSupportLinkURL() {
...
 
+  public String encodeUrlToUtf8(String urlStr) {
+    urlStr = urlStr.replaceAll("http://", "");
+    try {
+      urlStr = URLEncoder.encode(urlStr, "UTF-8");
+    } catch (UnsupportedEncodingException exp) {
+      LOGGER.error("Failed to encode url [" + urlStr + "] to utf-8", exp);
+    }
+    urlStr = "http://" + urlStr;
+    getContext().getResponse();
```
fpichler: why do you need to call getResponse?

fpichler (changes requested):

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1100,4 +1103,23 @@ public String getSupportLinkURL() {
     return url;
   }
 
+  public String encodeUrlToUtf8(String urlStr) {
+    urlStr = urlStr.replaceAll("http://", "");
```
fpichler: This is not functioning for other protocols e.g. https

fpichler (changes requested):

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1100,4 +1103,28 @@ public String getSupportLinkURL() {
     return url;
   }
 
+  public String encodeUrlToUtf8(String urlStr) {
+    String pattern = "://";
+    int findIndex = urlStr.indexOf(pattern);
+    if (findIndex > 0) {
+      String protocol = urlStr.substring(0, findIndex + pattern.length());
```
fpichler: This is not only the protocol. It is the protocoll and the delimiter. Hence call the variable e.g. urlPrefix

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1100,4 +1103,28 @@ public String getSupportLinkURL() {
...
+      String mainUrl = urlStr.substring(findIndex + pattern.length(), urlStr.length());
```
fpichler: you do not need to give urlStr.lenght() as ending parameter. Instead you can omit the ending parameter. As a default the Substring goes up to the end of the string

fpichler (changes requested):
please add unit tests

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1100,4 +1103,28 @@ public String getSupportLinkURL() {
     return url;
   }
 
+  public String encodeUrlToUtf8(String urlStr) {
```
fpichler: please move implementation to xxxService and only add a delegation in the ScriptService

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -1100,4 +1103,28 @@ public String getSupportLinkURL() {
...
+        urlStr = URLEncoder.encode(mainUrl, "UTF-8");
+      } catch (UnsupportedEncodingException exp) {
+        LOGGER.error("Failed to encode url [" + urlStr + "] to utf-8", exp);
+      }
+      urlStr = urlPrefix + urlStr;
+    }
+    return urlStr;
+  }
+
+  public void sendRedirect(String urlStr) {
```
fpichler: please move implementation to xxxService and only add a delegation in the ScriptService

fpichler (changes requested):

fpichler (approved):

PR #146, opened 2017-11-22
remove extracted celements-xwiki classes

fpichler:
https://synjira.atlassian.net/browse/CELDEV-302


msladek (changes requested):
LoadingTesting was moved to model, but not deleted in core

msladek (approved):

PR #148, opened 2017-12-01
only add line break if the closing tag is breaksFlow

fpichler:
https://synjira.atlassian.net/browse/CELDEV-600

### DIFF
--- a/src/main/java/com/celements/web/utils/Html2Text.java
+++ b/src/main/java/com/celements/web/utils/Html2Text.java
@@ -22,12 +22,15 @@ package com.celements.web.utils;
 import java.io.IOException;
 import java.io.Reader;
 
+import javax.swing.text.MutableAttributeSet;
+import javax.swing.text.html.HTML;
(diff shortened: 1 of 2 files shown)


src/main/java/com/celements/web/utils/Html2Text.java
```diff
@@ -22,12 +22,15 @@
...
+import javax.swing.text.MutableAttributeSet;
+import javax.swing.text.html.HTML;
 import javax.swing.text.html.HTMLEditorKit;
 import javax.swing.text.html.parser.ParserDelegator;
 
 public class Html2Text extends HTMLEditorKit.ParserCallback {
 
-  StringBuffer s;
+  private StringBuffer s;
+  private boolean tanglingReturn = false;
```
ebeutler: Not sure, but shouldn't it rather be 'dangling'?

ebeutler (dismissed):

ebeutler (approved):

PR #152, opened 2018-02-07
Move query package to celements-core to -model

msladek:
see [celements-xwiki PR#15](https://github.com/celements/celements-xwiki/pull/15)
needed for [CELDEV-635](https://synjira.atlassian.net/browse/CELDEV-635)


fpichler:
I think it is better to start a separat celements-query Package. Wdyt?

fpichler (approved):

PR #157, opened 2018-03-27
deleted logic for ClassCollections and MandatoryDoc in init

dkhasham:
https://synjira.atlassian.net/browse/CELDEV-650


src/main/java/com/celements/web/plugin/CelementsWebPlugin.java
```diff
@@ -101,12 +101,6 @@ public String getName() {
   public void init(XWikiContext context) {
     // TODO move to ApplicationReadyEvent listener after migration to xwiki 4
```
msladek: please also remove this TODO comment since there is nothing to move anymore. the log is fine

msladek (changes requested):

msladek (approved):

PR #171, opened 2018-06-21
EventListener#isDisabled fix for single entry

msladek:
avoids having to add a second empty line in properties file for single entries.


fpichler (changes requested):
can you please add a unit test for isDisabledInConfigSrc?

msladek:
Continued on https://github.com/celements/celements-core/pull/172

PR #182, opened 2018-11-28
Extension PresentationType/CellWriter

msladek:
https://synjira.atlassian.net/browse/CELDEV-763

### DIFF
--- a/src/main/java/com/celements/cells/ICellWriter.java
+++ b/src/main/java/com/celements/cells/ICellWriter.java
@@ -32,29 +32,33 @@ public interface ICellWriter {
    * @deprecated since 2.82 instead use openLevel(String, List<CellAttribute>)
    */
   @Deprecated
-  public void openLevel(@Nullable String tagName, @Nullable String idname,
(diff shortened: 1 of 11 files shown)


src/main/java/com/celements/cells/ICellWriter.java
```diff
@@ -32,29 +32,33 @@
...
-  public void openLevel(String idname, String cssClasses, String cssStyles);
+  void openLevel(String idname, String cssClasses, String cssStyles);
 
-  public void closeLevel();
+  void closeLevel();
 
-  public void clear();
+  void clear();
 
-  public String getAsString();
```
fpichler, line 48: Do not remove. Mark deprecated and add deprecated notice.

fpichler (changes requested):

src/main/java/com/celements/cells/ICellWriter.java
msladek: it's not removed but moved, see line 59

fpichler (approved):

PR #192, opened 2019-02-05
CELDEV-752 Improve ContextMenu loading Performance

msladek:
https://synjira.atlassian.net/browse/CELDEV-752

### DIFF
--- a/src/main/java/com/celements/navigation/service/TreeNodeService.java
+++ b/src/main/java/com/celements/navigation/service/TreeNodeService.java
@@ -478,8 +478,10 @@ public class TreeNodeService implements ITreeNodeService {
         EntityReference parent = getParentEntityRef(docRef);
         List<TreeNode> subMenuItems = getSubNodesForParent(parent, menuItem.getStringValue(
             "part_name"));
-        LOGGER.debug("getPrevMenuItem: {} subMenuItems found for parent '{}'. {}",
-            subMenuItems.size(), parent, Arrays.deepToString(subMenuItems.toArray()));
+        if (LOGGER.isDebugEnabled()) {
+          LOGGER.debug("getPrevMenuItem: {} subMenuItems found for parent '{}'. {}",
+              subMenuItems.size(), parent, Arrays.deepToString(subMenuItems.toArray()));
+        }
         int pos = getMenuItemPos(docRef, menuItem.getStringValue("part_name"));
         if (previous && (pos > 0)) {
           return subMenuItems.get(pos - 1);
--- a/src/main/java/com/celements/navigation/service/TreeNodeService.java
+++ b/src/main/java/com/celements/navigation/service/TreeNodeService.java
@@ -25,7 +25,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
-import java.util.Vector;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -389,7 +388,7 @@ public class TreeNodeService implements ITreeNodeService {
 
   @Override
   public Integer getMaxConfiguredNavigationLevel() {
-    List<BaseObject> navConfigObjects = new Vector<>();
+    List<BaseObject> navConfigObjects = new ArrayList<>();
     List<BaseObject> navConfigDocsObjs = getNavObjectsOnConfigDocs();
     if (navConfigDocsObjs != null) {
       navConfigObjects.addAll(navConfigDocsObjs);
@@ -412,14 +411,14 @@ public class TreeNodeService implements ITreeNodeService {
   }
 
   List<BaseObject> getNavObjectsFromLayout() {
-    List<BaseObject> navObjects = new Vector<>();
+    List<BaseObject> navObjects = new ArrayList<>();
     SpaceReference layoutRef = getPageLayoutCmd().getPageLayoutForCurrentDoc();
     if (layoutRef != null) {
       XObjectIterator objectIterator = new XObjectIterator(getContext());
-      List<String> layoutCellList = new Vector<>();
+      List<String> layoutCellList = new ArrayList<>();
       List<TreeNode> subNodesForParent = getSubNodesForParent(layoutRef, "");
       while (!subNodesForParent.isEmpty()) {
-        List<TreeNode> newSubNodes = new Vector<>();
+        List<TreeNode> newSubNodes = new ArrayList<>();
         for (TreeNode node : subNodesForParent) {
           List<TreeNode> subNodes = getSubNodesForParent(node.getDocumentReference(), "");
           newSubNodes.addAll(subNodes);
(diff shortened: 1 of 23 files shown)


src/main/java/com/celements/navigation/service/TreeNodeService.java
```diff
@@ -217,14 +216,28 @@ String getParentKey(EntityReference reference, boolean withDataBase) {
...
+        builder.add(node);
+      } else {
+        LOGGER.debug("fetchNodesForParent: omit [{}].", node);
+      }
+    }
+    return builder.build();
+  }
+
   /**
    * fetchNodesForParentKey
```
fpichler: please adjust javadoc

src/main/java/com/celements/navigation/service/TreeNodeService.java
```diff
@@ -217,14 +216,28 @@ String getParentKey(EntityReference reference, boolean withDataBase) {
     return getSubNodesForParent(entRef, filter);
   }
 
+  private List<TreeNode> fetchNodesForParent(EntityReference ref, INavFilter<?> filter) {
```
fpichler: filterNodesForParent instead of fetchNodesForParent

src/main/java/com/celements/velocity/CelementsVelocityService.java
```diff
@@ -0,0 +1,70 @@
...
+import org.xwiki.velocity.VelocityManager;
+import org.xwiki.velocity.XWikiVelocityException;
+
+import com.celements.model.context.ModelContext;
+import com.celements.model.util.ModelUtils;
+import com.google.common.base.Strings;
+import com.xpn.xwiki.doc.XWikiDocument;
+
+@Component
+public class CelementsVelocityService implements VelocityService {
```
fpichler: "Celements" is already in the package name. Call it VelocityService and VelocityServiceRole (for interface) or similar.

src/main/java/com/celements/velocity/CelementsVelocityService.java
```diff
@@ -0,0 +1,70 @@
...
+
+  private static Logger LOGGER = LoggerFactory.getLogger(CelementsVelocityService.class);
```
fpichler: make Loger final

src/main/java/com/celements/web/contextmenu/ContextMenuBuilder.java
```diff
@@ -120,53 +119,31 @@ public void stringEvent(String value) {
...
+        DocumentReference.class)).iter());
+    objs = objs.append(getObjectFetcher(refBuilder.with(getContext().getWikiRef()).build(
+        DocumentReference.class)).iter());
+    return objs;
   }
 
-  public void addElementsCMforClassNames(String jsonDictionary, XWikiContext context) {
+  public void addElementsCMforClassNames(String jsonDictionary) {
+    long time = System.currentTimeMillis();
     if ((jsonDictionary != null) && !"".equals(jsonDictionary)) {
```
fpichler: please refactor using !Strings.isNullOrEmpty

src/main/java/com/celements/web/contextmenu/ContextMenuItem.java
```diff
@@ -21,77 +21,49 @@
...
   private static final Logger LOGGER = LoggerFactory.getLogger(ContextMenuItem.class);
 
-  private String cmiLink;
-  private String cmiText;
-  private String cmiIcon;
-  private String shortcut;
-
-  private String elemId;
-  private String origElemId;
+  private final BaseObject menuItemObj;
```
fpichler: keep POJO setup. Move towards imputable objects to allow caching later on

fpichler (changes requested):

src/main/java/com/celements/velocity/CelementsVelocityService.java
msladek: we've dropped this interface syntax (e.g. `IArticleEngineRole`) some while ago and since use simple interface names, for a discussion see e.g. https://stackoverflow.com/a/2814831/1238689
we sometimes have trouble naming the implements though, `CelementsXService` or now `DefaultXService` is unsatisfactory. if you have a better idea let me know.

src/main/java/com/celements/web/contextmenu/ContextMenuItem.java
msladek: not sure I understand what you mean by "keep pojo setup". cloned obj though to ensure effective immutability.

fpichler: The target software architecture is a XWiki independent Context-Menu. Thus XWiki-classes should only be used to get read by the builder to fill in the ContextMenuItem values. Nevertheless We want to be able to create ContextMenuItem directly out ouf java without backing them by BaseObjects. Moreover the Java generated ContextMenuItem should only be generated as immutable objects and then kept in Memory. Thus the caching of the evaluated velocity scripts must be in different objects from different classes (mybe RenderedContextMenuItem). The RenderedContextMenuItem is only cachable in a user/request context. The ContextMenuItem will be cachable system-wide.

fpichler (changes requested):
please change the CelementsMenuItem implementation as mentioned in the conversion.

src/main/java/com/celements/web/contextmenu/ContextMenuItem.java
msladek: Thanks for clarification. I decoupled the `ContextMenuItem` from the `BaseObject` and introduced a constructor without it to also allow CMI creation without storage instances, according to your vision.

I agree that the current CMI is not suitable for system wide caching, I noted this in it's javadoc. At the time we're going to add a system wide CMI-cache we'll have to reconsider it's impact thoroughly (I remember the headache with the document cache) and potentially refactor the CMI to our needs. I feel venturing into this future feature is out of scope at this point.

msladek:
see changes and comment on the discussion

src/main/java/com/celements/web/contextmenu/ContextMenuItem.java
```diff
@@ -61,12 +69,19 @@ public ContextMenuItem(BaseObject menuItemObj, String elemId, XWikiContext conte
...
-    this.menuItemObj = (BaseObject) menuItemObj.clone();
-    this.renderCache = new HashMap<>();
+    this(elemId, menuItemObj.getStringValue("cmi_link"), menuItemObj.getStringValue("cmi_text"),
+        menuItemObj.getStringValue("cmi_icon"), menuItemObj.getStringValue("cmi_shortcut"));
+  }
+
+  public ContextMenuItem(String elemId, String link, String text, String icon, String shortcut) {
     this.origElemId = elemId;
-    this.elemIdParts = Arrays.asList(elemId.split(":"));
+    this.elemIdParts = Arrays.asList(elemId.split(":", -1));
```
fpichler, line 78: good fix 👍 
this change would have led to problems.

fpichler (approved):

PR #237, opened 2020-02-21
Cell render celOnEvent

msladek:
https://synjira.atlassian.net/browse/CELDEV-876

### DIFF
--- a/src/main/java/com/celements/cells/CellsClasses.java
+++ b/src/main/java/com/celements/cells/CellsClasses.java
@@ -59,13 +59,12 @@ public class CellsClasses extends AbstractClassCollection {
   @Deprecated
   public static final String CELLCLASS_IDNAME_FIELD = ICellsClassConfig.CELLCLASS_IDNAME_FIELD;
 
-  public CellsClasses() {
-  }
+  public CellsClasses() {}
 
   @Override
   protected void initClasses() throws XWikiException {
     LOGGER.debug("entering initClasses for database: " + getContext().getDatabase());
-    getCellClass();
+    // getCellClass();
     getPageLayoutPropertiesClass();
     getGroupCellClass();
     getPageDepCellConfigClass();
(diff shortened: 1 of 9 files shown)


src/main/resources/celements.hbm.xml
```diff
@@ -52,6 +52,9 @@
     <property name="css_styles" type="string">
       <column name="CEL_CSS_STYLES" />
     </property>
+    <property name="event_data_attr" type="string">
```
fpichler: The HTML5 definition does not give a max-length of an attribute value. HTML4 states 65635. Even though it is not obvious when to use a event-data-attr with that long data, I suggest to use `type="text"` instead. This makes obvious that there is now sensible limit.

src/main/java/com/celements/cells/CellsClasses.java
```diff
@@ -196,8 +197,9 @@ BaseClass getPageDepCellConfigClass() throws XWikiException {
   BaseClass getTranslationBoxCellConfigClass() throws XWikiException {
     XWikiDocument doc;
     boolean needsUpdate = false;
-    DocumentReference translationBoxCellConfigClassRef = cellsClassConfig.getTranslationBoxCellConfigClassRef(
-        getContext().getDatabase());
+    DocumentReference translationBoxCellConfigClassRef = cellsClassConfig
```
fpichler, line 198: I do not see how this should be more readable in three lines. please keep it on two.

src/main/java/com/celements/cells/CellsClasses.java
```diff
@@ -122,8 +122,9 @@ public DocumentReference getPageLayoutPropertiesClassRef(String wikiName) {
   BaseClass getPageLayoutPropertiesClass() throws XWikiException {
     XWikiDocument doc;
     boolean needsUpdate = false;
-    DocumentReference pageLayoutPropertiesClassRef = cellsClassConfig.getPageLayoutPropertiesClassRef(
-        getContext().getDatabase());
+    DocumentReference pageLayoutPropertiesClassRef = cellsClassConfig
```
fpichler, line 124: I do not see how this should be more readable in three lines. please keep it on two.

src/main/java/com/celements/cells/CellsClasses.java
```diff
@@ -59,13 +59,12 @@
...
 
-  public CellsClasses() {
-  }
+  public CellsClasses() {}
 
   @Override
   protected void initClasses() throws XWikiException {
     LOGGER.debug("entering initClasses for database: " + getContext().getDatabase());
-    getCellClass();
+    // getCellClass();
```
fpichler: why keep it as a comment? please delete.

fpichler (changes requested):

fpichler (approved):

PR #243, opened 2020-04-05
DateScriptService

msladek:
DateScriptService for handling java8+ date classes in velocity

### DIFF
--- a/src/main/java/com/celements/collections/service/StreamScriptService.java
+++ b/src/main/java/com/celements/collections/service/StreamScriptService.java
@@ -29,12 +29,16 @@ public class StreamScriptService implements ScriptService {
     return nullGuard(stream).collect(Collectors.toCollection(() -> collection));
   }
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/celements/collections/service/StreamScriptService.java
```diff
@@ -29,12 +29,16 @@
     return nullGuard(stream).collect(Collectors.toCollection(() -> collection));
   }
 
-  public String join(Stream<String> stream, String separator) {
```
fpichler, line 32: Can you add a "toString" for each element? Comming from velocity generics are not garantied.

fpichler (changes requested):

src/main/java/com/celements/collections/service/StreamScriptService.java
msladek: this comment references the old signature. please check the new one.

fpichler (approved):

PR #251, opened 2020-05-18
DocFormCommand: serial doc load&save

msladek:
https://synjira.atlassian.net/browse/CELDEV-899


src/main/java/com/celements/docform/DocFormCommand.java
```diff
@@ -0,0 +1,281 @@
...
+ * that they are correctly prepared for save. IMPORTANT: use exactly ONE instance of this
+ * class for each XWikiContext (each request).
+ *
+ * @author Fabian Pichler <fabian.pichler@synventis.com>
+ * @author Marc Sladek <marc@sladek.me>
+ */
+@Component
+@NotThreadSafe
+@InstantiationStrategy(PER_LOOKUP)
+public class DocFormCommand implements IDocForm {
```
fpichler, line 77: this is no Command anymore. Thus please rename to DocFormService or DocFormDefault or similar.

fpichler (changes requested):

src/main/java/com/celements/docform/DocFormCommand.java
msladek: The class still conforms to the command pattern, this behaviour hasn't changed with this PR.

The object instantiation is now handled by the `ComponentManager` with `@InstantiationStrategy(PER_LOOKUP)` for dependency management, but that doesn't qualify the class as a Service imho.

fpichler: ok. I see your point: https://alvinalexander.com/java/java-command-design-pattern-in-java-examples/

fpichler (dismissed):

msladek:
resolved merge conflicts

fpichler (approved):

PR #264, opened 2020-11-08
AddTranslationCommand remove setTranslations & unit test fixes

msladek:
https://synjira.atlassian.net/browse/CELDEV-702

### DIFF
--- a/src/main/java/com/celements/docform/DocFormCommand.java
+++ b/src/main/java/com/celements/docform/DocFormCommand.java
@@ -19,10 +19,12 @@
  */
 package com.celements.docform;
 
+import static com.celements.logging.LogLevel.*;
 import static com.celements.logging.LogUtils.*;
 import static com.celements.model.util.References.*;
 import static com.google.common.base.Predicates.*;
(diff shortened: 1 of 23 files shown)


src/main/java/com/celements/docform/DocFormCommand.java
```diff
@@ -196,7 +196,7 @@ XWikiDocument updateDocFromParam(XWikiDocument xdoc, XWikiDocument tdoc,
 
   private XWikiDocument setDocField(XWikiDocument tdoc, DocFormRequestParam param) {
     DocFormRequestKey key = param.getKey();
-    ClassField<String> field = xDocClassDef.getField(key.getFieldName(), String.class).orNull();
+    ClassField<String> field = xDocClassDef.getField(key.getFieldName(), String.class).orElse(null);
```
fpichler: why at all converting the optional to null? 'field' is only used twice in the following lines. This could be easily done with the optional instead of a null check.

fpichler (changes requested):

fpichler (approved):

PR #268, opened 2021-01-10
update parent pom

fpichler:


fpichler:
upgrade not working with old jenkins

PR #297, opened 2022-02-02
CELDEV-1006 - refactor AttachmentURLCommand to ResourceUrlService

fpichler:
https://synjira.atlassian.net/browse/CELDEV-1006


src/main/resources/META-INF/components.txt
```diff
@@ -160,3 +160,4 @@ com.celements.cells.classes.PageLayoutPropertiesClass
 com.celements.cells.classes.GroupCellClass
 com.celements.cells.classes.PageDepCellConfigClass
 com.celements.cells.classes.TranslationBoxCellConfigClass
+com.celements.ressource_url.RessourceUrlService
```
msladek:
```suggestion
com.celements.ressource.url.RessourceUrlService
```

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    return false;
+  }
+
+  private String getAction(Optional<String> action) {
+    if (action.isPresent()) {
+      return action.get();
+    }
+    return getDefaultAction();
+  }
```
msladek: superfluous method, the clients can simply use `action.orElseGet(this::getDefaultAction)`

src/main/java/com/celements/web/plugin/cmd/AttachmentURLCommand.java
```diff
@@ -20,85 +20,187 @@
...
     }
     return url;
   }
 
+  private String getAction(Optional<String> action) {
+    if (action.isPresent()) {
+      return action.get();
+    }
+    return getDefaultAction();
+  }
```
msladek: superfluous method, the clients can simply use `action.orElseGet(this::getDefaultAction)`

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+    return false;
+  }
+
+  @Override
+  public boolean isOnDiskLink(@Nullable String link) {
+    if (link != null) {
+      return ON_DISK_LINK_PATTERN.get().matcher(link.trim()).matches();
+    }
+    return false;
+  }
```
msladek: This kind of logic is better encapsulated in some kind of object `ResourceLink`. The object may expose these methods:
- `URL getUrl()`
- `@NotNull String getResourceName()`
- `@NotNull DocumentReference getDocRef()`
- `boolean isOnDisk()`

The service may then simply provide a method `Optional<ResourceLink> parseResourceLink(@Nullable String link)` and other methods expect a `ResourceLink` object instead of a `String`

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+  private String getAction(Optional<String> action) {
+    if (action.isPresent()) {
+      return action.get();
+    }
+    return getDefaultAction();
+  }
+
+  private String getDefaultAction() {
+    return context.getXWikiContext().getWiki().getXWikiPreference("celdefaultAttAction",
+        "celements.attachmenturl.defaultaction", "file", context.getXWikiContext());
```
msladek: why not from a `ConfigurationSource` with xwiki.cfg as fallback?

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  @Override
+  @NotEmpty
+  public String getRessourceURLPrefix() {
```
msladek: naming camelCase:
```suggestion
  public String getRessourceUrlPrefix() {
```
change everywhere accordingly

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+    return getRessourceURLPrefix(getDefaultAction());
+  }
+
+  @Override
+  @NotEmpty
+  public String getRessourceURLPrefix(@NotEmpty String action) {
+    XWikiURLFactory urlf = context.getXWikiContext().getURLFactory();
+    return urlf.createResourceURL("", true, context.getXWikiContext()).toString().replace("/skin/",
+        "/" + action + "/");
```
msladek: formatting/naming
```suggestion
    XWikiURLFactory urlFactory = ;
    return context.getXWikiContext().getURLFactory()
        .createResourceURL("", true, context.getXWikiContext()).toString()
        .replace("/skin/", "/" + action + "/");
```

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+
+  @Override
+  @NotEmpty
+  public String getRessourceURLPrefix() {
+    return getRessourceURLPrefix(getDefaultAction());
+  }
+
+  @Override
+  @NotEmpty
+  public String getRessourceURLPrefix(@NotEmpty String action) {
```
msladek: fast fail on non-empty

src/main/java/com/celements/ressource_url/UrlRessourceNotExistException.java
```diff
@@ -0,0 +1,29 @@
+package com.celements.ressource_url;
+
+import javax.validation.constraints.NotNull;
+
+public class UrlRessourceNotExistException extends Exception {
```
msladek: naming:
- resource with one `s`
- inconsistent with service naming: `ResourceUrlNotExistsException`

src/main/java/com/celements/ressource_url/RessourceUrlService.java
```diff
@@ -0,0 +1,185 @@
...
+      LOGGER.error("Error getting attachment URL for doc '{}' and file {}", getPageDocRef(link),
+          attName, exp);
+      throw new UrlRessourceNotExistException(link);
+    } catch (AttachmentNotExistsException anee) {
+      LOGGER.info("Attachment not found for link [{}] and action [{}]", link, action, anee);
+      throw new UrlRessourceNotExistException(link);
+    }
+  }
+
+  private String createOnDiskUrl(String link, Optional<String> action) {
```
msladek: for all methods ending with Url, better return an actual `java.net.URL` instance. see (and perhaps use) `com.celements.web.service.UrlService.createURIBuilder(EntityReference)` for building URI/URLs

msladek (changes requested):

src/main/java/com/celements/ressource_url/RessourceUrlService.java
fpichler: introduced FileReference

src/main/java/com/celements/ressource_url/RessourceUrlService.java
fpichler: done

src/main/java/com/celements/filebase/references/FileReference.java
```diff
@@ -0,0 +1,202 @@
...
+
+  @NotThreadSafe
+  public static final class Builder {
+
+    private static final String ATTACHMENT_LINK_REGEX = "([\\w\\-]*:)?([\\w\\-]*\\.[\\w\\-]*){1};.*";
+    private static final Supplier<Pattern> ATTACHMENT_LINK_PATTERN = Suppliers
+        .memoize(() -> Pattern.compile(ATTACHMENT_LINK_REGEX));
+    private static final String ON_DISK_LINK_REGEX = "^:[/\\w\\-\\.]*";
+    private static final Supplier<Pattern> ON_DISK_LINK_PATTERN = Suppliers
+        .memoize(() -> Pattern.compile(ON_DISK_LINK_REGEX));
```
msladek: why do we need a memoize here? The patterns can be intialised statically. Also the regex's can be inlined:
```suggestion
  private static final Pattern ATTACHMENT_LINK_PATTERN = Pattern.compile(
      "([\\w\\-]*:)?([\\w\\-]*\\.[\\w\\-]*){1};.*");
  private static final Pattern ON_DISK_LINK_PATTERN = Pattern.compile("^:[/\\w\\-\\.]*");
```

src/main/java/com/celements/filebase/references/FileReference.java
```diff
@@ -0,0 +1,202 @@
...
+        && Objects.equals(((FileReference) obj).fullPath, this.fullPath);
+  }
+
+  @Override
+  public String toString() {
+    return "FileReference [name=" + name + ", type=" + type + ", docRef=" + docRef + ", fullPath="
+        + fullPath + "]";
+  }
+
+  public static Builder of(@NotEmpty String link) {
```
msladek: since it returns a Builder please move it to the inner class
`var myBuilder = FileReference.Builder.of(link);`

src/main/java/com/celements/filebase/references/FileReferenceType.java
```diff
@@ -0,0 +1,7 @@
+package com.celements.filebase.references;
+
+public enum FileReferenceType {
```
msladek: since this is strongly coupled as the type of `FileReference`'s, why not as public inner enum?
`var type = FileReference.Type.ON_DISK;`

src/main/java/com/celements/filebase/uri/FileUriScriptService.java
```diff
@@ -0,0 +1,105 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.xwiki.component.annotation.Component;
+import org.xwiki.component.annotation.Requirement;
+import org.xwiki.script.service.ScriptService;
+
+import com.celements.filebase.references.FileReference;
+import com.google.common.base.Strings;
+
+import groovy.lang.Singleton;
```
msladek: wrong import. add `groovy.lang.*` to your IDE auto complete filters ([wiki](https://intern.synventis.com/SynWiki/Eclipse-Java-Code-Style)).

src/main/java/com/celements/filebase/references/FileReference.java
```diff
@@ -0,0 +1,202 @@
...
+
+  }
+
+  private final String name;
+  private final FileReferenceType type;
+  private final DocumentReference docRef;
+  private final String fullPath;
+  private final String queryString;
+
+  public FileReference(Builder builder) {
```
msladek:
```suggestion
  private FileReference(Builder builder) {
```

src/main/java/com/celements/filebase/references/FileReference.java
```diff
@@ -0,0 +1,202 @@
...
+    this.name = builder.name;
+    this.type = builder.type;
+    this.fullPath = builder.fullPath;
+    this.docRef = builder.docRef;
+    this.queryString = builder.queryString;
+  }
+
+  public String getName() {
```
msladek, line 136: all non-primitive methods in this class: do they return nullable/notnull?

src/main/java/com/celements/filebase/uri/FileUriServiceRole.java
```diff
@@ -0,0 +1,26 @@
...
+  @NotNull
+  UriBuilder createFileUri(@NotNull FileReference fileRef, @NotNull Optional<String> action,
+      @NotNull Optional<String> queryString) throws FileNotExistException;
+
+  @NotNull
+  UriBuilder getFileUriPrefix(@NotNull Optional<String> action);
+
+  @NotNull
+  UriBuilder createAbsoluteFileUri(@NotNull FileReference fileRef, @NotNull Optional<String> action,
+      Optional<String> queryString);
```
msladek: all methods in this service: [RSPEC-3553 "Optional" should not be used for parameters](https://rules.sonarsource.com/java/RSPEC-3553)

src/main/java/com/celements/filebase/uri/FileUriScriptService.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+    return UriBuilder.fromPath("");
+  }
+
+  @NotNull
+  public UriBuilder createFileUrl(@Nullable FileReference fileRef, @Nullable String action,
+      @Nullable String queryString) {
+    if (fileRef != null) {
+      try {
+        return fileUriService.createFileUri(fileRef, Optional.ofNullable(action),
```
msladek, line 60: why is the service method called `...Uri`, the SS name `...Url` but both return a `UriBuilder`? a few thoughts:
1. velocity cannot handle the exceptions thrown by `UriBuilder#build` and `URI#toURL`. So if we actually handle URLs here, the java logic should capture any `MalformedURLException`.
2. Afaik the `FileUriService` actually builds URLs, so why not return `URL` objects? This would encapsulate the handling of URL malformation inside the service and make it's usage more robust. If a client wants to change the URL, it can always create a `UriBuilder` from it.

src/main/java/com/celements/filebase/uri/FileUriService.java
```diff
@@ -0,0 +1,160 @@
...
+      uriBuilder = createOnDiskUri(fileRef, action);
+    } else {
+      uriBuilder = fileRef.getUri();
+    }
+    return addContextUri(uriBuilder);
+  }
+
+  UriBuilder addContextUri(UriBuilder uriBuilder) {
+    Optional<XWikiDocument> currentDoc = context.getCurrentDoc().toJavaUtil();
+    if (currentDoc.isPresent() && uriBuilder.toString().startsWith("?")) {
```
msladek, line 124: I don't think checking the toString of the builder is a good idea since we have no guarantee as to how it's implemented apart from being a "textual" representation. perhaps you want to check for no path?

src/main/java/com/celements/filebase/uri/FileUriService.java
```diff
@@ -0,0 +1,160 @@
...
+      uriBuilder.replacePath(currentDoc.get().getURL("view", context.getXWikiContext()));
```
msladek, line 125: I think this does nothing since `replacePath` isn't inplace. perhaps write a test for this method.

src/main/java/com/celements/filebase/uri/FileUriService.java
```diff
@@ -0,0 +1,160 @@
...
+    return uriBuilder;
+  }
+
+  private UriBuilder createAttachmentUri(@NotNull FileReference fileRef, Optional<String> action)
+      throws FileNotExistException {
+    String attName = fileRef.getName();
+    try {
+      XWikiDocument doc = modelAccess.getDocument(fileRef.getDocRef());
+      XWikiAttachment att = attachmentSrv.getAttachmentNameEqual(doc, attName);
+
```
msladek, line 136:
```suggestion
```

src/main/java/com/celements/filebase/uri/FileUriService.java
```diff
@@ -0,0 +1,160 @@
...
+      throw new FileNotExistException(fileRef);
+    }
+  }
+
+  UriBuilder createOnDiskUri(@NotNull FileReference fileRef, Optional<String> action) {
+    String path = fileRef.getFullPath().trim().substring(1);
+    UriBuilder uri = UriBuilder.fromPath(
+        context.getXWikiContext().getWiki().getSkinFile(path, true, context.getXWikiContext())
+            .replace("/skin/",
+                "/" + action.orElse(getDefaultAction()) + "/"));
```
msladek, line 155:
```suggestion
            .replace("/skin/", "/" + action.orElse(getDefaultAction()) + "/"));
```

src/main/java/com/celements/filebase/uri/FileUriService.java
```diff
@@ -0,0 +1,160 @@
...
+    uri.queryParam("version", lastStartupTimeStamp.getFileModificationDate(path));
+    return uri;
```
msladek, line 157:
```suggestion
    return uri.queryParam("version", lastStartupTimeStamp.getFileModificationDate(path));
```

src/main/java/com/celements/web/service/CelementsWebScriptService.java
```diff
@@ -563,7 +588,8 @@ public String getDefaultLanguage(String spaceName) {
   }
 
   private String getDeletedDocsHql(String orderby, boolean hideOverwritten) {
-    String deletedDocsHql = "select distinct ddoc.fullName" + " from XWikiDeletedDocument as ddoc";
+    String deletedDocsHql = "select distinct ddoc.fullName"
+        + " from XWikiDeletedDocument as ddoc";
```
msladek, line 592:
```suggestion
    String deletedDocsHql = "select distinct ddoc.fullName from XWikiDeletedDocument as ddoc";
```

msladek (changes requested):

src/main/java/com/celements/filebase/references/FileReference.java
fpichler: included build() call, now returning FileReference. Because after call to of the FileReference is fully quallyfied

src/main/java/com/celements/filebase/uri/FileUriScriptService.java
fpichler: to 1. There is a #try statement for velocity
to 2. Nope URL must be fully qualified. UriBuilder can handle e.g. "/folder/to/file.js" which is not fully quallified and returns a correct toString()

src/main/java/com/celements/filebase/uri/FileUriService.java
fpichler: Well, there is no other way to check the state of the URI-Builder. Furthermore "toString" is the only way to NOT get a "IlligaleArgumentException" for not fully qualified urls.

src/main/java/com/celements/filebase/uri/FileUriScriptService.java
msladek: 1. I was referring to calling `UriBuilder#build` and `URI#toURL` within velocity. the try/catch here won't prevent `MalformedURLException` thrown outside of the try/catch.
2. afaict `UriBuilder#toString` isn't documented/guaranteed behaviour, see [javadoc](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.25/jersey/index.html?javax/ws/rs/core/UriBuilder.html).

src/main/java/com/celements/filebase/uri/FileUriService.java
msladek: I see, it's really unfortunate that the UriBuilder interface doesn't have getters for the URI parts.
afaict `UriBuilder#toString` isn't documented/guaranteed behaviour, see [javadoc](https://eclipse-ee4j.github.io/jersey.github.io/apidocs/2.25/jersey/index.html?javax/ws/rs/core/UriBuilder.html).

src/main/java/com/celements/filebase/uri/FileUriScriptService.java
msladek: 1. I was referring to calling `UriBuilder#build` and `URI#toURL` within velocity. the try/catch here won't prevent `MalformedURLException` thrown outside of the try/catch.
2. it can be misleading calling the SS method names `...Url...` if they don't return at minimum Url's but incomplete Uri's. The service methods are named `...Uri...`. Why this inconsistency?

PR #298, opened 2022-02-05
pom.xml: fix shared tests dependency

msladek:
https://synjira.atlassian.net/browse/CELDEV-1002

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,13 +31,6 @@
   <version>5.4-SNAPSHOT</version>
   <description>Celements Core</description>
   <dependencies>
-    <!-- Add here all your dependencies -->
-    <dependency>
-      <groupId>com.celements</groupId>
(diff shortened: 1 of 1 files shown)


fpichler (approved):

pom.xml
```diff
@@ -65,24 +63,27 @@
...
-          <artifactId>dom3-xml-apis</artifactId>
-        </exclusion>
-      </exclusions>
+    </dependency>
+
+    <!-- testing -->
+    <dependency>
+      <groupId>com.celements</groupId>
+      <artifactId>celements-shared-tests</artifactId>
+      <version>5.1-SNAPSHOT</version>
```
fpichler, line 72: is already added in the parent "components" pom.

fpichler (changes requested):

PR #306, opened 2022-05-19
CELDEV-846 - add getRteJsonConfigField to RTEConfigScriptService

fpichler:
https://synjira.atlassian.net/browse/CELDEV-846


src/main/java/com/celements/rteConfig/RTEConfigScriptService.java
```diff
@@ -77,6 +78,16 @@ public String getRTEConfigField(@NotNull String name) {
     return "";
   }
 
+  @NotNull
+  public JsonBuilder getRteJsonConfigField(@NotNull String name) {
```
msladek: allow nullable parameter

msladek (changes requested):

msladek (approved):

PR #309, opened 2022-05-24
CELDEV-1046 - add support for javascript module files (mime-type and script tag)

fpichler:
https://synjira.atlassian.net/browse/CELDEV-1046

### DIFF
--- a/src/main/java/com/celements/javascript/JsFileEntry.java
+++ b/src/main/java/com/celements/javascript/JsFileEntry.java
@@ -68,7 +68,7 @@ public final class JsFileEntry extends ObjectBean {
   }
 
   public boolean isModule() {
-    String fileName = getFilepath().split("[?$]")[0];
+    String fileName = UriBuilder.fromUri(getFilepath()).build().getPath();
     return fileName.endsWith(".mjs");
   }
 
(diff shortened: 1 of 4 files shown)


ebeutler (dismissed):

src/main/java/com/celements/javascript/JsFileEntry.java
```diff
@@ -67,6 +67,11 @@ public boolean isValid() {
     return !jsFileUrl.isEmpty();
   }
 
+  public boolean isModule() {
+    String fileName = getFilepath().split("[?$]")[0];
```
msladek: since `getFilePath` misleadingly doesn't return a path (as defined by [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3)) but may also contain a query, it's better to use an existing parser instead of custom parsing: 
```suggestion
    String fileName = UriBuilder.fromUri(getFilepath()).build().getPath();
```
This also works for empty input strings.

src/test/java/com/celements/javascript/JsFileEntryTest.java
```diff
@@ -96,6 +96,30 @@ public void test_isValid_valid() {
...
+    jsFileEntry.setFilepath("/space/doc/attachment.mjs");
+    assertTrue(jsFileEntry.isModule());
+  }
+
+  @Test
+  public void test_isModule_yes_search() {
+    jsFileEntry.setFilepath("/space/doc/attachment.mjs?version=asdf");
+    assertTrue(jsFileEntry.isModule());
+  }
+
```
msladek, line 128: also add a test without setting filepath (empty string)

msladek (changes requested):

msladek (approved):

PR #311, opened 2022-06-07
CELDEV-1046 - fix JsFileEntry#isModule for onDisk paths

msladek:
https://synjira.atlassian.net/browse/CELDEV-1046

solves test failure


fpichler:
It is the test which is wrong not the implementation. The mocking on line 73 is wrong. _getAttachmentURL_ must return a real path. Use the following lines for the mocking instead.
```
    String file = ":celJS/prototype.js";
    String filePath = "/file/celJS/prototype.js?version=12344";
    expect(attUrlCmd.getAttachmentURL(eq(file), same(context))).andReturn(filePath).once();
```

fpichler (changes requested):
please fix the test not the implementation

fpichler (approved):

PR #313, opened 2022-08-14
DateScriptService#format/parseISO

msladek:
https://synjira.atlassian.net/browse/CELDEV-1057


fpichler:
instead: use velocity date tool.

msladek:
The DateScriptService is intended as a Java 8 Date/Time API replacement for the VelocityTools DateTool. For formatting/parsing it uses the DateTimeFormatter instead of SimpleDateFormat.
The DateTool doesn't provide ISO 8601 formatting capabilities, at least not in the version included.

fpichler (approved):

PR #324, opened 2022-12-06
XDocumentFieldComparator

msladek:
will be used for sorting rows in https://synjira.atlassian.net/browse/CELDEV-1073

### DIFF
--- a/src/main/java/com/celements/cells/CellRenderStrategy.java
+++ b/src/main/java/com/celements/cells/CellRenderStrategy.java
@@ -23,6 +23,7 @@ import static com.celements.model.util.ReferenceSerializationMode.*;
 
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Stream;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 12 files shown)


src/main/java/com/celements/cells/CellRenderStrategy.java
```diff
@@ -58,7 +58,7 @@
 
 public class CellRenderStrategy implements IRenderStrategy {
 
-  public static final String EXEC_CTX_KEY = "celements.globalvalues.cell";
+  public static final String EXEC_CTX_KEY = "celements.cell";
```
fpichler, line 63: is it correct to remove "globalvalues." from the constant if considering the same constant in CelementsWebScriptService.java:102 still showing "globalvalues."?

fpichler (changes requested):

src/main/java/com/celements/cells/CellRenderStrategy.java
msladek: I've removed `globalvalues` since I don't want to expose the rendering scope to velocity, mostly because we're now adding XWikiDocument instances to it.
In order to not break legacy solutions (which indeed manipulate `celements.globalvalues.cell.number` from velocity) I've added a fallback for it where necessary.

fpichler: 👍

fpichler (approved):

PR #327, opened 2022-12-19
CELDEV-1076 - replace old js lazyLoad in TabEditor with cel-lazy-load-js

fpichler:
https://synjira.atlassian.net/browse/CELDEV-1076

### DIFF
--- a/src/main/java/com/celements/javascript/JsFileEntry.java
+++ b/src/main/java/com/celements/javascript/JsFileEntry.java
@@ -4,12 +4,11 @@ import static com.google.common.base.Preconditions.*;
 
 import java.util.Objects;
 import java.util.Optional;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 import javax.validation.constraints.NotNull;
+import javax.ws.rs.core.UriBuilder;
 
 import com.celements.model.object.ObjectBean;
 import com.google.common.base.Strings;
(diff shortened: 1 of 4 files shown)


src/main/java/com/celements/javascript/JsFileEntry.java
```diff
@@ -72,7 +76,11 @@ public boolean isValid() {
   }
 
   public boolean isModule() {
-    String fileName = UriBuilder.fromUri(getFilepath()).build().getPath();
+    String fileName = getFilepath();
+    Matcher matcher = Pattern.compile(".*/([^/?]+).*").matcher(getFilepath());
```
msladek: I'm generally wary of writing manual parsers when there's already a library for it. If it's only about the leading `:` we add for marking onDisk paths, why not just use a `replaceFirst("^:", "")` before handing it to the `UriBuilder`? I think it would be better readable and less error prone then using a regex for path/query matching.

src/main/java/com/celements/web/plugin/cmd/ExternalJavaScriptFilesCommand.java
```diff
@@ -213,13 +212,8 @@ public String getLazyLoadTag(@NotNull ExtJsFileParameter extJsFileParams) {
...
       @Nullable AttachmentURLCommand attUrlCmdMock) {
-    final JsonBuilder jsonBuilder = new JsonBuilder();
-    jsonBuilder.openDictionary();
-    jsonBuilder.addProperty("fullURL", generateUrl(extJsFileParams, attUrlCmdMock));
-    jsonBuilder.addProperty("initLoad", true);
-    jsonBuilder.closeDictionary();
-    return "<span class='cel_lazyloadJS' style='display: none;'>" + jsonBuilder.getJSON()
-        + "</span>";
+    return "<cel-lazy-load-js src=\"" + generateUrl(extJsFileParams, attUrlCmdMock)
+        + "\" loadMode=\"" + extJsFileParams.getLoadMode() + "\"><!-- test --></cel-lazy-load-js>";
```
msladek: remove the test comment? also consider using the `DivWriter`.

msladek (changes requested):

msladek (approved):

PR #331, opened 2023-03-05
VelocityScriptService

msladek:


fpichler:
Can you please add a Jira-Issue where it is needed for? Please consider too, that there are already similar script service commands: $services.celementsweb.renderDocument

msladek:
I didn't know about `renderDocument`, only `renderCelementsDocument` which adds surrounding divs. This is what I was looking for as a replacement of `$xwiki.includeForm`, thanks.

PR #339, opened 2023-04-13
UrlScriptService catch Exceptions

msladek:
https://synjira.atlassian.net/browse/CELDEV-1106

### DIFF
--- a/src/main/java/com/celements/web/service/UrlScriptService.java
+++ b/src/main/java/com/celements/web/service/UrlScriptService.java
@@ -2,6 +2,8 @@ package com.celements.web.service;
 
 import static com.google.common.base.MoreObjects.*;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/celements/web/service/UrlScriptService.java
```diff
@@ -14,62 +16,78 @@
...
 
   @Requirement
   private ModelContext context;
 
   public String getURL() {
-    return urlService.getURL(getCurrentReference());
+    return getURL(null);
   }
 
   public String getURL(EntityReference ref) {
```
fpichler: please add nullable annotations to the parameters

fpichler (changes requested):

fpichler (approved):

PR #356, opened 2023-08-09
CELDEV-1167 - add new EnsureConsistentUserStateListener

cbspichler:
https://synjira.atlassian.net/browse/CELDEV-1167


src/main/java/com/celements/auth/user/listener/EnsureConsistentUserStateListener.java
```diff
@@ -0,0 +1,104 @@
...
+import com.celements.observation.save.SaveEventOperation;
+import com.celements.observation.save.object.ObjectEvent;
+import com.celements.pagetype.classes.PageTypeClass;
+import com.celements.rights.access.EAccessLevel;
+import com.celements.web.classes.oldcore.XWikiRightsClass;
+import com.celements.web.classes.oldcore.XWikiUsersClass;
+import com.xpn.xwiki.doc.XWikiDocument;
+import com.xpn.xwiki.user.api.XWikiUser;
+
+@Component
```
msladek: this class implements the `ComponentRole` `EventListener` thus we need to set a name:
```suggestion
@Component(EnsureConsistentUserStateListener.NAME)
```
if no name is set it defaults to the name `default` which could cause issues due to duplicate namings. if it weren't a  `ComponentRole` it'd default to the class FQN as the name. in these cases you don't need to set an explicit name.

src/main/java/com/celements/auth/user/listener/EnsureConsistentUserStateListener.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public List<Event> getEvents() {
+    return List.of(new ObjectEvent(SaveEventOperation.CREATING, XWikiUsersClass.CLASS_REF));
+  }
+
+  @Override
+  protected void onEventInternal(@NotNull Event event, @NotNull XWikiDocument source, Object data) {
+    LOGGER.trace(
+        "onObjectEvent in ensureConsistentUserState for source {} and data {}", source,
+        data);
```
msladek:
```suggestion
    LOGGER.trace("onObjectEvent in ensureConsistentUserState for source {} and data {}",
        source, data);
```

msladek (changes requested):

msladek (approved):

PR #359, opened 2023-08-24
HTML Rendering Errors

msladek:
https://synjira.atlassian.net/browse/CELDEV-1179

### DIFF
--- a/src/main/java/com/celements/javascript/JsFileEntry.java
+++ b/src/main/java/com/celements/javascript/JsFileEntry.java
@@ -1,5 +1,6 @@
 package com.celements.javascript;
 
+import static com.celements.javascript.JsLoadMode.*;
 import static com.google.common.base.Preconditions.*;
 
 import java.util.Objects;
@@ -72,7 +73,7 @@ public final class JsFileEntry extends ObjectBean {
 
   @NotNull
   public JsLoadMode getLoadMode() {
-    if (isModule()) {
+    if (isModule() && (loadMode == SYNC)) {
       return JsLoadMode.DEFER;
     }
     return loadMode;
--- a/src/main/java/com/celements/web/plugin/cmd/ExternalJavaScriptFilesCommand.java
+++ b/src/main/java/com/celements/web/plugin/cmd/ExternalJavaScriptFilesCommand.java
@@ -19,6 +19,7 @@
  */
 package com.celements.web.plugin.cmd;
 
+import static com.celements.javascript.JsLoadMode.*;
 import static com.google.common.base.Preconditions.*;
 
 import java.util.Collections;
@@ -50,7 +51,6 @@ import com.celements.convert.bean.XObjectBeanConverter;
 import com.celements.javascript.ExtJsFileParameter;
 import com.celements.javascript.JavaScriptExternalFilesClass;
 import com.celements.javascript.JsFileEntry;
-import com.celements.javascript.JsLoadMode;
 import com.celements.model.access.IModelAccessFacade;
 import com.celements.model.access.exception.DocumentNotExistsException;
 import com.celements.model.context.ModelContext;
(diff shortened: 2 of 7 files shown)


fpichler:
Do not see the “Error” for module-scripts with defer. [<script>: The Script element - HTML: HyperText Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) clearly states: “The defer attribute has no effect on [module scripts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) — they defer by default.” Thus, it has no effect but is no ERROR.

src/main/java/com/celements/web/plugin/cmd/ExternalJavaScriptFilesCommand.java
```diff
@@ -291,14 +291,13 @@ void injectDisplayAll(boolean displayedAll) {
...
-    return "<script" + ((jsFile.getLoadMode() != JsLoadMode.SYNC)
-        ? " " + jsFile.getLoadMode().toString().toLowerCase()
-        : "") + " type=\""
-        + (jsFile.isModule()
-            ? "module"
-            : "text/javascript")
-        + "\" src=\""
-        + StringEscapeUtils.escapeHtml(jsFile.getFilepath()) + "\"></script>";
+    var loadMode = Optional.ofNullable(jsFile.getLoadMode())
+        .filter(mode -> !jsFile.isModule() && (mode != JsLoadMode.SYNC));
```
fpichler: This is wrongly filtering loadMode=async for module which is valid and therfore must be supported. See [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script)

-> obviously a unit test for type=module and loadMode=async is missing :-) therefore please add one.

fpichler (changes requested):

msladek:
The [W3C Validator](https://validator.w3.org/#validate_by_input) lists the error:
`Error: A script element with a defer attribute must not have a type attribute with the value module.`
for this input:
```
<!DOCTYPE html>
<html>
  <head>
    <title>title</title>
  </head>
  <body>
    <script defer src="./static/js/main.js" type="module"></script>
  </body>
</html>
```

src/main/java/com/celements/web/plugin/cmd/ExternalJavaScriptFilesCommand.java
msladek: this cannot happen because of the implementation of `JsFileEntry.getLoadMode()`, where we force defer for all modules.
```
  public JsLoadMode getLoadMode() {
    if (isModule()) {
      return JsLoadMode.DEFER;
    }
    return loadMode;
  }
```

msladek: I've changed the check to `(isModule() && (loadMode == SYNC))`

fpichler:
jep, W3C Validator is broken. I cannot find this restriction in the https://www.w3.org/TR/2011/WD-html5-author-20110809/the-script-element.html ... nvm

src/main/java/com/celements/javascript/JsFileEntry.java
```diff
@@ -72,7 +73,7 @@ public String getFilepath() {
 
   @NotNull
   public JsLoadMode getLoadMode() {
-    if (isModule()) {
+    if (isModule() && (loadMode == SYNC)) {
```
fpichler, line 76: 👍

fpichler (approved):

msladek:
The W3C validator is not broken.
> Module scripts may specify the async attribute, but must not specify the defer attribute.

https://html.spec.whatwg.org/multipage/scripting.html#attr-script-defer

PR #363, opened 2023-09-28
CELDEV-1190 - add support for local iso date time

fpichler:
https://synjira.atlassian.net/browse/CELDEV-1190


src/main/java/com/celements/collections/service/DateScriptService.java
```diff
@@ -161,6 +162,25 @@ public ZonedDateTime parseISO(String text) {
     return parse(PATTERN_ISO, ZoneId.of("UTC"), text);
   }
 
+  public ZonedDateTime parseLocalISO(String text, String defaultTime, ZoneId zone) {
+    if (!text.contains("T")) {
+      text += "T" + defaultTime;
+    }
+    return LocalDateTime.parse(text).atZone(zone);
```
msladek: avoid NPEs from null arguments and catch possible parsing exceptions.

msladek (changes requested):

msladek (dismissed):

msladek (approved):

PR #366, opened 2023-10-14
CELDEV-1198 - ContentDynLoad must pass through request parameters

fpichler:
https://synjira.atlassian.net/browse/CELDEV-1198

### DIFF
--- a/src/main/java/com/celements/navigation/presentation/RenderedContentDynLoadPresentationType.java
+++ b/src/main/java/com/celements/navigation/presentation/RenderedContentDynLoadPresentationType.java
@@ -1,13 +1,19 @@
 package com.celements.navigation.presentation;
 
+import java.net.URL;
+import java.util.List;
+import java.util.Optional;
+import java.util.StringJoiner;
+
+import javax.inject.Inject;
 import javax.validation.constraints.NotNull;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/celements/navigation/presentation/RenderedContentDynLoadPresentationType.java
```diff
@@ -1,13 +1,19 @@
...
+import java.util.StringJoiner;
+
+import javax.inject.Inject;
 import javax.validation.constraints.NotNull;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.xwiki.component.annotation.Component;
-import org.xwiki.component.annotation.Requirement;
+import org.springframework.stereotype.Component;
```
msladek, line 14: also remove from components.txt

msladek (changes requested):

src/main/java/com/celements/navigation/presentation/RenderedContentDynLoadPresentationType.java
fpichler: done

msladek (approved):

PR #368, opened 2023-10-15
CELDEV-1199 - "version" queryparam missing for External JS Files

msladek:
https://synjira.atlassian.net/browse/CELDEV-1199


src/main/java/com/celements/web/plugin/cmd/AttachmentURLCommand.java
```diff
@@ -21,6 +21,7 @@
 
 import static com.celements.common.MoreOptional.*;
 import static com.celements.execution.XWikiExecutionProp.*;
+import static org.python.google.common.base.Strings.*;
```
fpichler: Do you really need "python" Strings not `com.google.common.base.Strings` instead?

fpichler (changes requested):

fpichler (approved):

PR #383, opened 2024-12-07
CELDEV-1260 WikiCreatedEventListener order

msladek:
https://synjira.atlassian.net/browse/CELDEV-1260

### DIFF
--- a/src/main/java/com/celements/mandatory/XWikiXWikiPreferences.java
+++ b/src/main/java/com/celements/mandatory/XWikiXWikiPreferences.java
@@ -100,7 +100,7 @@ public class XWikiXWikiPreferences extends AbstractMandatoryDocument {
         .createFirstIfNotExists();
     boolean dirty = false;
     dirty |= setStringValue(prefsObj, "title", "Celements");
-    dirty |= setStringValue(prefsObj, "skin", "cellskin");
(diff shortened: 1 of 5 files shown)


src/main/java/com/celements/mandatory/XWikiXWikiPreferences.java
```diff
@@ -100,7 +100,7 @@ private boolean checkWikiPreferences(XWikiDocument wikiPrefDoc,
         .createFirstIfNotExists();
     boolean dirty = false;
     dirty |= setStringValue(prefsObj, "title", "Celements");
-    dirty |= setStringValue(prefsObj, "skin", "celskin");
+    dirty |= setStringValue(prefsObj, "skin", "cellskin");
```
fpichler:
```suggestion
    dirty |= setStringValue(prefsObj, "skin", "");
```

the default skin in xwiki.cfg is already cellskin:
`#-# The default skin to use when there's no value specified in the wiki preferences document. Note that the default
#-# wiki already specifies a skin, so this setting is only valid for empty wikis.
xwiki.defaultskin=cellskin`
I therefore prefere to not force it additionally in the XWiki/XWikiPreferences

fpichler (approved):

PR #389, opened 2025-03-09
CELDEV-1271 - start component celements-mail-sender / split from celements-core

fpichler:
https://synjira.atlassian.net/browse/CELDEV-1271


src/main/java/com/celements/web/plugin/api/CelementsWebPluginApi.java
cbspichler, line 1: Can we be sure that the deleted methods aren't used in scripts that aren't on disk?

cbspichler (approved):

src/main/java/com/celements/web/plugin/api/CelementsWebPluginApi.java
fpichler: I have searched for them on my workspace. Is this "sure" enough?

PR #407, opened 2026-03-12
just formatting celements-core repo

fpichler:
only formatting all resources with eclipse-format.xml


src/main/java/com/celements/metatag/enums/opengraph/EOpenGraph.java
```diff
@@ -10,19 +10,17 @@
...
+      "og:image"), OPENGRAPH_URL(
+          "og:url"), OPENGRAPH_OPTIONAL_AUDIO("og:audio"), OPENGRAPH_OPTIONAL_DESCRIPTION(
+              "og:description"), OPENGRAPH_OPTIONAL_DETERMINER(
+                  "og:determiner"), OPENGRAPH_OPTIONAL_LOCALE(
+                      "og:locale "), OPENGRAPH_OPTIONAL_LOCALE_ALTERNATE(
+                          "og:locale:alternate"), OPENGRAPH_OPTIONAL_SITENAME(
+                              "og:site_name"), OPENGRAPH_OPTIONAL_VIDEO(
+                                  "og:video"), OPENGRAPH_OPTIONAL_IMAGE_WIDTH(
+                                      "og:image:width"), OPENGRAPH_OPTIONAL_IMAGE_HEIGHT(
+                                          "og:image:height");
```
msladek, line 23: not all is better here

msladek (commented):
I can't see the actual formatting changes because most of the diff is just `CRLF` to `LF` conversions. Are these line-ending changes really needed?
If so, please make a distinct PR with only the line-ending conversions so the actual formatting becomes visible here.

PR #416, opened 2026-07-02
Consolidate mandatory XWiki Rights/Preferences

msladek:
https://synjira.atlassian.net/browse/PROGON-1423

### DIFF
--- a/src/main/java/com/celements/mandatory/XWikiXWikiRights.java
+++ b/src/main/java/com/celements/mandatory/XWikiXWikiRights.java
@@ -19,9 +19,6 @@
  */
 package com.celements.mandatory;
 
-import static com.celements.logging.LogUtils.*;
-
-import java.util.Arrays;
 import java.util.List;
 
 import org.slf4j.Logger;
@@ -31,7 +28,11 @@ import org.xwiki.component.annotation.Requirement;
 import org.xwiki.model.reference.DocumentReference;
 
 import com.celements.model.classes.ClassDefinition;
+import com.celements.model.object.xwiki.XWikiObjectEditor;
+import com.celements.model.object.xwiki.XWikiObjectFetcher;
+import com.celements.model.reference.RefBuilder;
 import com.celements.web.classes.oldcore.XWikiGlobalRightsClass;
+import com.xpn.xwiki.XWikiConstant;
 import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.doc.XWikiDocument;
 import com.xpn.xwiki.objects.BaseObject;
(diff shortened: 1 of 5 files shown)


src/main/java/com/celements/mandatory/XWikiXWikiRights.java
```diff
@@ -75,39 +79,41 @@ protected boolean checkDocumentsMain(XWikiDocument doc) throws XWikiException {
...
-      LOGGER.debug("XWikiPreferences missing access rights fixed for [{}]", getWiki());
-      return true;
+    boolean dirty = false;
+    dirty |= checkGlobalRights(wikiPrefDoc, "XWiki.ContentEditorsGroup", "edit,delete,undelete");
+    dirty |= checkGlobalRights(wikiPrefDoc, "XWiki.XWikiAdminGroup",
+        "admin,edit,comment,delete,undelete,register");
+    return dirty;
+  }
+
+  protected boolean checkGlobalRights(XWikiDocument wikiPrefDoc, String groupFN, String levels) {
```
cbspichler: naming is weird. it checks for global rights apparently. so I'd expect true if the exist and false if they don't. but it returns false if they exist. And if they don't exist, they are created and then it returns true.

src/main/java/com/celements/mandatory/XWikiXWikiRights.java
```diff
@@ -75,39 +79,41 @@ protected boolean checkDocumentsMain(XWikiDocument doc) throws XWikiException {
   }
 
   boolean checkAccessRights(XWikiDocument wikiPrefDoc) throws XWikiException {
```
cbspichler: naming is weird. this doesn't just check access rights on something.

src/main/java/com/celements/mandatory/XWikiXWikiRights.java
```diff
@@ -31,22 +28,26 @@
...
+import com.celements.model.object.xwiki.XWikiObjectEditor;
+import com.celements.model.object.xwiki.XWikiObjectFetcher;
+import com.celements.model.reference.RefBuilder;
 import com.celements.web.classes.oldcore.XWikiGlobalRightsClass;
+import com.xpn.xwiki.XWikiConstant;
 import com.xpn.xwiki.XWikiException;
 import com.xpn.xwiki.doc.XWikiDocument;
 import com.xpn.xwiki.objects.BaseObject;
 
 @Component("celements.mandatory.wikirights")
```
cbspichler, line 39: could be turned into a Spring bean if you want

cbspichler (approved):

